PR #42, opened 2018-09-29
Chat backend

FlorianVeaux:
Simple backend for chat messages. There is a single room to fetch messages from.

Two custom routes are added to retrieve a batch of messages at once.

The first one 'retrieve_up_to' will fetch a given amount of messages up to a given id. Can be used to retrieve the last XX messages when loading the page and to retrieve older messages.

The second one 'retrieve_from' will fetch new messages after a given id. Can be called very often to check for new messages to display.


abocquet:
Why do checks fail ?

FlorianVeaux:
https://github.com/Mines-Paristech-Students/Portail-des-eleves/pull/43 is fixing the tests (a package was not installed correctly). This branch is not using this code yet but the master is, I'll rebase on master to verify the tests are working correctly

PR #46, opened 2018-10-29
Adds the widget logic and a chat widget

FlorianVeaux:
Adds a `widget` logic in Angular and adds a first (functional) chat widget on the left side.


abocquet:
The widget crashes when there is not message in the database 
image

I tried to quick fix it with a try ... except but it does not work either (even though the error changes)

FlorianVeaux:
If there is no messages in the DB this caused the issue you mentioned. It has been fixed, and master merged into the branch.

Merging...

PR #59, opened 2018-12-17
17wan fat/polls

abocquet:


polls/admin.py
```diff
@@ -0,0 +1,46 @@
...
+            }
+        ),
+        (
+            'Auteur',
+            {
+                'fields': ['user']
+            }
+        ),
+        (
+            'Validation',
```
FlorianVeaux: We never decided if the language of the administration panel should be French or English, but I think this is a reasonable choice to make it easier for new administrators.

polls/models.py
```diff
@@ -0,0 +1,91 @@
+from django.db import models
+from django.utils import timezone
+
+from authentication.models import User
+
+
+class Poll(models.Model):
+    # The question of the poll.
+    question = models.CharField(
+        'question',
```
FlorianVeaux: I struggled to find the meaning of this positional argument in the docs, https://docs.djangoproject.com/en/2.0/topics/db/models/#verbose-field-names

Maybe we can replace this line by 
```suggestion
        verbose_name='question',
```

So it's easier to understand

polls/models.py
```diff
@@ -0,0 +1,91 @@
...
+    )
+    state = models.CharField(
+        'état',
+        max_length=9,
+        choices=STATES,
+        default='REVIEWING',
+    )
+
+    # The date of publication.
+    publication_date_time = models.DateTimeField(
```
FlorianVeaux: I think that having polls starting at midnight everyday and being available for one day is something a lot of people care about. Having an unique DateField to replace `publication_date_time` and `publication_end_date_time` will be easier to manage.

polls/models.py
```diff
@@ -0,0 +1,91 @@
...
+    def is_published(self):
+        """
+        Return True if the poll is currently published, that is publication_date_time <= now() <=
+        publication_date_time + duration, and its state is ACCEPTED.
+        """
+        now = timezone.now()
+        if self.publication_date_time is None or self.publication_end_date_time is None:
+            return False
+        else:
+            return self.state == 'ACCEPTED' and self.publication_date_time <= now <= self.publication_end_date_time
```
FlorianVeaux: By following the previous advice, this function wont be needed because:
- We rely on publication date to publish a poll
- No need to have the 'ACCEPTED' state, the publication date should exist only if the poll has been previously accepted by an admin. The publication date can be either manually set by an admin, or if one day there is no poll for the given day, one can pick a previously accepted poll and set its date as the current's date

polls/models.py
```diff
@@ -0,0 +1,91 @@
...
+
+    is_published.short_description = 'publié ?'
+    is_published.boolean = True
+    is_published.admin_order_field = 'publication_date_time'
+
+
+class Choice(models.Model):
+    # The text of the choice.
+    text = models.CharField(
```
FlorianVeaux: It decouples the choices and the poll. Nice, it will allow polls with 3, 4 or more choices to choose from

polls/views/poll.py
```diff
@@ -0,0 +1,90 @@
...
+
+    permission_classes = (permissions.IsAuthenticated,)
+
+    def get(self, request, id):
+        poll = get_object_or_404(Poll, id=id)
+
+        if request.user.is_admin or poll.user == request.user:
+            serialized_poll = WholePollSerializer(poll, context={'request': request})
+        else:  # the user is not an admin and the poll was not submitted by them.
+            if not poll.is_published():
```
FlorianVeaux: poll.is_published() returns True if and only if the given poll is currently published, right ?
We will want users to have access to older polls too

polls/admin.py
```diff
@@ -0,0 +1,46 @@
...
+            {
+                'fields': ['state', 'publication_date_time', 'publication_end_date_time', 'admin_comment']
+            }
+        ),
+    ]
+
+
+@admin.register(Choice)
```
FlorianVeaux: I don't think we want to add the Choice to the admin panel. This will add some noise that is not necessary as we should be able to edit the choices values within the poll modification form.

backend/tests.py
```diff
@@ -0,0 +1,60 @@
+from django.urls import reverse
+
+from rest_framework.test import APITestCase
+
+from authentication.models import User
+
+
+class BackendTestCase(APITestCase):
```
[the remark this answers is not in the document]
FlorianVeaux: I got it, this way we can juste extend `BackendTestCase` in all other tests

PR #61, opened 2018-12-30
Abocquet/associations

abocquet:
Basic user management is done but some work remains : 
- allow people to have title inside the associations (ex: "president", "treasurer"...)
- right management


associations/serializers/association.py
```diff
@@ -37,6 +36,41 @@ class Meta:
         model = Association
         fields = ('id', 'name', 'logo', 'pages', 'marketplace', 'library', 'groups')
 
-    def create(self, validated_data):
-        instance = Association.objects.create(**validated_data)
+    def update(self, instance, validated_data):
+        groups_data = validated_data.pop('groups')
+        instance = super().update(instance, validated_data)
```
FlorianVeaux: Please comment all of this, it's not really easy to follow

associations/serializers/association.py
```diff
@@ -37,6 +36,41 @@ class Meta:
...
+
+        ids = set()
+
+        for group_data in groups_data:
+            id = dict(group_data)["id"]
+
+            try:
+                group = Group.objects.get(pk=id)
+            except Exception: # help requested : too broad catch
```
FlorianVeaux: `Group.DoesNotExist` ;-)

associations/serializers/association.py
```diff
@@ -37,6 +36,41 @@ class Meta:
...
+                    ids.add(group.id)
+
+            if not serialized.is_valid():
+                raise Exception(serialized.errors)
+
+        for group in instance.groups.all():
+            if group.id not in ids:
+                group.delete()
+
+        instance.save()
```
FlorianVeaux: I feel like all of this is too complicated, I don't really see why we should allow the API to create missing groups.
We need to talk about all this

associations/tests/test_groups.py
```diff
@@ -0,0 +1,238 @@
...
+        """
+
+        self.logout()
+
+        url = reverse('token_obtain_pair')
+        data = {'id': username, 'password': password}
+        return self.client.post(url, data, format='json')
+
+    def test_can_edit_group(self):
+        tests = [
```
FlorianVeaux: Can we load this from outside the code ?

frontend/src/app/associations/association-members/association-members.component.html
```diff
@@ -11,29 +11,24 @@
...
+                    </div>
+                    <div class="card-body">
+                        <div class="row">
+                            <div class="col-lg-2 col-md-4 col-6 text-center" *ngFor="let member of group.members_detail">
+                                <a routerLink="/user/{{ member.id }}">
+                                    <profile-photo [user]="member"></profile-photo>
+                                    <p class="mt-2">{{ member.first_name }} {{ member.last_name }}</p>
+                                </a>
+                            </div>
+                        </div>
```
FlorianVeaux: 👍

frontend/src/app/associations/association-members/association-members.component.ts
```diff
@@ -31,13 +33,29 @@ export class AssociationMembersComponent implements OnInit {
...
         const id = this.route.snapshot.paramMap.get('id');
-        this.api.get("associations/" + id + "/").subscribe(
-            association => this.association = association,
-            error => this.error = error.message
-        );
 
         this.api.get('users/').subscribe(
-            users => this.users = users,
+            users => {
+                let users_index = {};
```
FlorianVeaux: Not a big fan of all that because the users request is gonna be huge as it gets all data from all users. We will fetch way too much data for what will be used. Second I don't see where association.members is used (at least in its updated form), can you point to it?

I would populate the association field directly from the api request without changing members. If there is not enough information in what is returned from the API then change the API.
Second I would fetch only the list of usernames and use that in the ng-select element.

FlorianVeaux:
I made some changes, ready to merge

PR #75, opened 2019-02-20
Adds a notification service

FlorianVeaux:
Also cleans and update a few angular packages


frontend/src/app/home/home.component.ts
```diff
@@ -14,14 +15,12 @@ export class HomeComponent implements OnInit {
     is_settings_displayed: boolean;
     widgets: {"name": string, "displayed": boolean}[]
     associations: any[];
-    status_message_html: string;
 
-    constructor(private _apiService: ApiService) {
+    constructor(private _apiService: ApiService, private notifier: NotifierService) {
```
14alarro: _apiService mais notifier : pourquoi pas la même convention (sur l'underscore comme sur le suffixe "Service") ?

FlorianVeaux: The leading underscore looks like bad practice (https://angular.io/guide/styleguide#style-03-04), we should follow the same convention everywhere though

14alarro: Agreed, I would write apiService and notifierService

FlorianVeaux: 👍

PR #79, opened 2019-02-27
Abocquet/facebook

abocquet:


frontend/src/app/facebook/facebook.component.ts
```diff
@@ -15,14 +15,23 @@ import {ApiService} from "../api.service";
...
     p = 0; // The current page
 
 
-    constructor(private apiService: ApiService) {
+    constructor(private api: ApiService) {
     }
 
     ngOnInit() {
-        this.$users = this.apiService.getUsers();
+        this.$users = this.api.getUsers();
```
14alarro: I think we should keep the "Service" and "Components" suffixes, it is easier to understand

PR #93, opened 2019-09-07
Added tests for associations.models.library.Library.

Piwakk:
Only very general tests have been written so far (basic CRUD on a
library page).


associations/tests/base_test.py
```diff
@@ -27,18 +27,34 @@ def login(self, username, password='password'):
         return self.client.post(url, data, format='json')
 
     def delete(self, endpoint=''):
+        """Shortcut for self.client.delete('/api/v1/' + endpoint)."""
```
abocquet: I bit redundant with the code

associations/tests/base_test.py
```diff
@@ -27,18 +27,34 @@ def login(self, username, password='password'):
...
         return self.client.delete(self.api_base + endpoint)
 
     def get(self, endpoint=''):
+        """Shortcut for self.client.get('/api/v1/' + endpoint, format='json', content_type='application/json')."""
```
abocquet: Same here

associations/tests/base_test.py
```diff
@@ -27,18 +27,34 @@ def login(self, username, password='password'):
...
         return self.client.get(self.api_base + endpoint, format='json', content_type='application/json')
 
     def patch(self, endpoint='', data=''):
+        """
```
abocquet: Same here

associations/tests/base_test.py
```diff
@@ -27,18 +27,34 @@ def login(self, username, password='password'):
...
+                format='json',
+                content_type='application/json').
+        """
         return self.client.patch(self.api_base + endpoint,
                                  data=data,
                                  format='json',
                                  content_type='application/json')
 
     def post(self, endpoint='', data=''):
+        """
```
abocquet: Same here

associations/tests/test_library.py
```diff
@@ -2,111 +2,246 @@
...
+            self.assertTrue(Library.objects.get(pk='bd-tek').enabled)
 
-        self.login('17admin')  # Just global administrator, not library administrator.
-        res = self.delete('library/bd-tek/')
-        self.assertEqual(res.status_code, 403)
-        self.assertTrue(Library.objects.filter(pk='bd-tek').exists())
+    def test_if_library_admin_then_can_update_library(self):
+        self.login('17library_bd-tek')
+        res = self.patch('library/bd-tek/', data={'enabled': 'false'})
+        self.assertIn(res.status_code, [200, 204])
```
abocquet: Why are there 2 codes ? Isn't it deterministic ?

Piwakk: I think it is, but I'm not sure about which one is sent by the server…

abocquet: Choose one, we'll fix it when needed

PR #95, opened 2019-09-14
17wf/testsMarketplace

Piwakk:


associations/views/marketplace.py
```diff
@@ -1,147 +1,202 @@
...
     queryset = Product.objects.all()
     serializer_class = ProductSerializer
-    permission_classes = (CanManageMarketplace,)
+    permission_classes = (IfMarketplaceAdminThenCRUDElseR,
+                          IfMarketplaceEnabledThenCRUDElseRAndMarketplaceAdminOnlyCRUD,)
+
     filter_backends = (filters.SearchFilter,)
     search_fields = ("name",)
 
+    def get_queryset(self):
```
abocquet: This function has no body

Piwakk: That's because it's nobody ~

Piwakk: Actually it has a body:
```python
"""The user has access to the products coming from every enabled marketplace and to the products of every
        marketplace she is a library administrator of."""

        # The library for which the user is library administrator.
        marketplaces = [role.association.marketplace for role in self.request.user.roles.all() if role.marketplace]

        return Product.objects.filter(Q(marketplace__enabled=True) | Q(marketplace__in=marketplaces))
```

Piwakk: Actually it does have a body:
```python
"""The user has access to the products coming from every enabled marketplace and to the products of every
        marketplace she is a library administrator of."""

        # The library for which the user is library administrator.
        marketplaces = [role.association.marketplace for role in self.request.user.roles.all() if role.marketplace]

        return Product.objects.filter(Q(marketplace__enabled=True) | Q(marketplace__in=marketplaces))
```

PR #96, opened 2019-09-15
17wf/testsUser

Piwakk:


authentication/models/user.py
```diff
@@ -37,48 +39,37 @@ def _create_user(self, id, first_name, last_name, email, password, birthday, pro
...
+    first_name = models.CharField(max_length=50)
+    last_name = models.CharField(max_length=50)
     nickname = models.CharField(max_length=128, blank=True, default="")
-    birthday = models.DateField(null=True, verbose_name="date de naissance")
-
-    # Contact the person
-    phone = models.CharField(max_length=15, blank=True, verbose_name="numéro de téléphone")
+    birthday = models.DateField(null=True)
+    email = models.EmailField(max_length=160, unique=True)
+    promo = models.IntegerField(verbose_name="Promotion",
```
abocquet: do we leave `verbose_name` ?

authentication/models/user.py
```diff
@@ -37,48 +39,37 @@ def _create_user(self, id, first_name, last_name, email, password, birthday, pro
...
-    city_of_origin = models.CharField(max_length=128, blank=True, help_text="ville d'origine")
+                               help_text="Address outside the Meuh. Blank if the User is not PAM.")
+    city_of_origin = models.CharField(max_length=128, blank=True)
 
-    # Cursus
+    # Education.
     option = models.CharField(max_length=128, blank=True)
     is_ast = models.BooleanField(default=False)
     is_isupfere = models.BooleanField(default=False)
     is_in_gapyear = models.BooleanField(default=False)
```
abocquet: Maybe change the name of this variable to keep it even after the perso has come back from gapyear ?

authentication/models/user.py
```diff
@@ -90,22 +81,30 @@ class Meta:
...
-
     @property
     def is_staff(self):
-        """Is the user a member of staff?"""
-        # Simplest possible answer: All admins are staff
         return self.is_admin
 
+    @property
+    def is_in_first_year(self):
+        """Return True iff the User is in her first year at the school, depending on her promotion."""
```
abocquet: we got only female users ?

authentication/models/user.py
```diff
@@ -90,22 +81,30 @@ class Meta:
...
+        if today.month >= 8:
+            # Between August and December, the new students have the same school year as the current year.
+            return self.promo == today.year % 100
+        else:
+            # Between January and July, the school year of the new students is the current year minus one.
+            return self.promo == (today.year % 100) - 1
+
+    @property
+    def show_secret(self):
+        """Return True iff the User can be shown… the SECRET."""
```
abocquet: Spoiler alert, please be a bit less explicit

authentication/permissions.py
```diff
@@ -0,0 +1,28 @@
+from rest_framework.permissions import BasePermission, SAFE_METHODS
+
+
+class ProfilePermission(BasePermission):
+    """
+               | Own profile | Another user |
+        Admin  | CRUD        | CRUD         |
+        Simple | RU          | R            |
```
abocquet: So clear ❤️

authentication/serializers/user.py
```diff
@@ -7,16 +7,26 @@
 from authentication.token import Token
 
 
+class UpdateUserSerializer(serializers.ModelSerializer):
```
abocquet: Maybe simple use `to_representation` instead of two serializers ?

authentication/views/profile.py
```diff
@@ -1,50 +1,96 @@
...
+                                    'minesparent', ]
+    """The profile fields editable by a simple (non-admin) user."""
 
-class UserViewSet(viewsets.ModelViewSet):
-    """
-    API endpoint that allows user to be viewed or edited.
-    """
     queryset = User.objects.all()
     serializer_class = UserSerializer
+    permission_classes = [ProfilePermission]
```
abocquet: Use tuple instead (non mutable)

authentication/views/profile.py
```diff
@@ -1,50 +1,96 @@
...
+            return UpdateUserSerializer
+        elif self.action in ('list', ):
             return UserShortSerializer
         else:
             return UserSerializer
+
+    def update(self, request, *args, **kwargs):
+        """
+            Can X update Y? \n
+            X / Y       | Herself                               | Another user | \n
```
abocquet: Only 👩 ?

authentication/models/user.py
Piwakk: I left it because “promo” is not an English word

authentication/views/profile.py
Piwakk: I always thought you had to use feminine in English when you didn't know the gender :D

Piwakk: I always thought you had to use feminine in English when you didn't know the gender :D

authentication/models/user.py
Piwakk: Problem is, if we change it to “did_a_gapyear”, we cannot know if he or she is currently in a gapyear or not (because of people who repeat years).

authentication/views/profile.py
abocquet: Nope, use `them` ;-)

Piwakk: Looks like it's better to use _they_/_their/…
https://english.stackexchange.com/questions/48/is-there-a-correct-gender-neutral-singular-pronoun-his-vs-her-vs-their

authentication/models/user.py
abocquet: This raises the more general issue of people who repeated a year. They might be excluded from votes for instance

Piwakk: I would be in favour of creating something like a new Model AcademicYear with three fields:
* the user;
* the beginning year of the academic year (such as 2017 for 2017-2018);
* the academic year (1A, 2A, Césure…).

It would also help categorizing the AST as people who must not have access to the secret…

Piwakk: BTW, I think we should remove the field is_ast, and replace it by student_type (or something like this), with options like 'AST', 'EV', 'ISUPFERE', 'IC'…

abocquet: The `AcademicYear` could be directly integrated to `User`. Explicitly stating the academic year sounds good, although I'm not sure to get the point if having the beginning of the year

Piwakk: OK, so a `list<tuple<int, str>>` such as `[(2017, '1A'), (2018, '1A'), (2019, '2A'), (2020, 'GAP_YEAR'), (2021, '3A')]`?
But how do we enforce, in Django, that the second member of the `tuple` is always something in `1A`, `2A`, …?

By the beginning year I mean, instead of storing 2017-2018 which is a string, only store 2017, which is a nice integer.

Piwakk: Or maybe you meant why do we need to store the academic year (2017-2018 or 2017 or whatever) as we have the “promotion”?
Well, maybe some people will leave the school or whatever. And your promotion may change. Explicit is better than implicit, as they say. ;)

abocquet: Do we really need to keep track of the whole education history of the person ? Wouldn't just be current one be enough ?

Piwakk: Well, we want to know if it's their first year at school or not (for elections, secret…)

abocquet: In that case you'll only need to know if they are currently in 1A, 2A..., not what were their history

Piwakk: Imagine someone who repeated their 1A.
Or an AST, who directly goes to 2A.
In both cases you won't know

abocquet: Let's keep 
- `cursus_type`: "IC", "EV"...
- `year`: 1A, 2A, gap year, 3A
- `year_of_entry`: 2017, ...

authentication/tests/test_profile.py
```diff
@@ -89,34 +88,66 @@ def test_cannot_create_user(self):
...
-        res = self.patch('users/17simple/', data=self.limited_edit_user_data)
+        res = self.patch('users/17simple/', data=self.allowed_edits)
         self.assertStatusCode(res, 200)
 
         user = User.objects.get(pk='17simple')
-        for field in self.limited_edit_user_data:
+        for field in self.allowed_edits:
             if field in ['roommate', 'minesparent']:
-                self.assertEqual([u.id for u in user.__getattribute__(field).all()], self.limited_edit_user_data[field])
+                self.assertEqual([u.id for u in user.__getattribute__(field).all()], self.allowed_edits[field])
```
abocquet: maybe user the function `getattr` instead of `. __getattribute__ `

authentication/tests/test_profile.py
```diff
@@ -89,34 +88,66 @@ def test_cannot_create_user(self):
...
             else:
-                self.assertEqual(user.__getattribute__(field), self.full_edit_user_data[field])
+                self.assertEqual(user.__getattribute__(field), self.allowed_edits[field])
+
+    def test_if_admin_then_cannot_edit_other_user_profile_with_disallowed_edits(self):
+        self.login('17admin')
+        user_before = User.objects.get(pk='17simple')
+        res = self.patch('users/17simple/', data=self.not_allowed_edits)
+        self.assertStatusCode(res, 200)
```
abocquet: If the edits are not allowed, it shouldn't return a `200` code, should it ?

authentication/tests/test_profile.py
```diff
@@ -89,34 +88,66 @@ def test_cannot_create_user(self):
...
-                self.assertEqual([u.id for u in user.__getattribute__(field).all()], self.full_edit_user_data[field])
+                self.assertEqual([u.id for u in user.__getattribute__(field).all()], self.allowed_edits[field])
             else:
-                self.assertEqual(user.__getattribute__(field), self.full_edit_user_data[field])
+                self.assertEqual(user.__getattribute__(field), self.allowed_edits[field])
+
+    def test_if_admin_then_cannot_edit_other_user_profile_with_disallowed_edits(self):
+        self.login('17admin')
+        user_before = User.objects.get(pk='17simple')
+        res = self.patch('users/17simple/', data=self.not_allowed_edits)
```
abocquet: Maybe test the `non_allowed_edits` one by one to make sure every one of them aren't editable

authentication/tests/test_profile.py
Piwakk: Actually no, because the DRF serializer just ignores the read_only fields.
So, we end up with blank update data, which still makes a successful update

PR #98, opened 2019-09-17
17wf/events

Piwakk:


associations/serializers/events.py
```diff
@@ -0,0 +1,39 @@
...
+        if 'starts_at' in self.initial_data and 'ends_at' in self.initial_data:
+            if self.initial_data['starts_at'] >= self.initial_data['ends_at']:
+                self._errors = {'field_errors': 'field starts_at is not consistent with field ends_at.'}
+
+        if self._errors and raise_exception:
+            raise ValidationError(self._errors)
+
+        return super(EventSerializer, self).is_valid(raise_exception)
+
+    def update(self, instance, validated_data):
```
abocquet: This function is a bit redundant with `is_valid`. I think that `is_valid` must be called before `update` for serializers

associations/tests/events/test_events.py
```diff
@@ -0,0 +1,128 @@
...
+
+    def test_if_event_admin_then_can_create_event(self):
+        self.login('17events_biero')  # Event administrator.
+        res = self.create(association_id='biero', data=self.event)
+        self.assertStatusCode(res, 201)
+        self.assertTrue(Event.objects.filter(name=self.event['name']).exists())
+        self.assertEqual(Event.objects.get(name=self.event['name']).description, self.event['description'])
+        self.assertEqual(Event.objects.get(name=self.event['name']).place, self.event['place'])
+        self.assertEqual(Event.objects.get(name=self.event['name']).starts_at, self.event['starts_at'])
+        self.assertEqual(Event.objects.get(name=self.event['name']).ends_at, self.event['ends_at'])
```
abocquet: Maybe avoid querying the event each time for faster tests

associations/tests/events/test_events.py
```diff
@@ -0,0 +1,128 @@
...
+            self.assertStatusCode(res, 403)
+            self.assertEqual(Event.objects.get(pk=data['pk']).name, 'Biéro')
+
+    def test_if_event_admin_then_can_update_event(self):
+        self.login('17events_biero')
+        data = {'pk': 1, 'name': 'Biéroloose', 'description': 'Personne'}
+        res = self.update(data['pk'], data=data, association_id='biero')
+        self.assertStatusCode(res, 200)
+        self.assertEqual(Event.objects.get(pk=data['pk']).name, data['name'])
+        self.assertEqual(Event.objects.get(pk=data['pk']).description, data['description'])
```
abocquet: same

associations/serializers/events.py
Piwakk: Unfortunately, `is_valid` does not give access to the updated instance. So we either do that verification in `update` or in the view. I think both are the same (but at least, here, we are being provided the instance).

PR #100, opened 2019-09-18
17wf/tests page

Piwakk:


associations/tests/page/base_test_page.py
```diff
@@ -0,0 +1,54 @@
...
+"""Same as ALL_USERS, but with 17page_biero removed."""
+
+ALL_USERS_EXCEPT_PAGE_ADMIN = [user for user in ALL_USERS if 'page' not in user]
+"""Same as ALL_USERS, but with 17page* removed."""
+
+ALL_USERS_EXCEPT_ADMIN = [user for user in ALL_USERS if 'admin' not in user]
+"""Same as ALL_USERS, but with all the association administrators removed."""
+
+
+class BasePageTestCase(BaseTestCase):
```
abocquet: Many redundancies with the News tests, maybe there is a way to refactor it

Piwakk: Yeah. As we previously discussed, a way to do it would be to merge `News` and `Page` into a `Page` model, only adding a `type` field which would store the… type of page.

Piwakk: This will be done… one day. For now, let's keep the tests as they are ~

PR #103, opened 2019-09-19
17wf/elections

Piwakk:


associations/models/election.py
```diff
@@ -0,0 +1,66 @@
...
+        Represents a choice in an election.
+    """
+
+    id = models.AutoField(primary_key=True)
+
+    election = models.ForeignKey(Election, on_delete=models.CASCADE, related_name='choices')
+    name = models.CharField(max_length=200)
+
+
+class Vote(models.Model):
```
abocquet: Vote is a bit ambiguous, as it can be the time when you vote or the bulletin

abocquet: Maybe `Election` -> `ElectionManager` or `AssociationElection` or `VogteBureau`and `Vote` -> `Election` ?

authentication/models/user.py
```diff
@@ -135,8 +135,15 @@ def years_since_entry(self):
...
-        for role in self.roles.all():
-            if role.association == association:
-                return role
+    def get_role(self, association=None, association_pk=None):
+        q = None
+
+        if association_pk is not None:
+            q = self.roles.filter(association_id=association_pk)
+        elif association is not None:
+            q = self.roles.filter(association=association)
```
abocquet: I beleive you can use `self.roles.filter(association=association)` if `association` is either an `Association` or the association id

associations/models/election.py
Piwakk: Well, _election_ does not mean anything else than _élection_ in French (https://www.wordreference.com/enfr/election). But you are right, the English _vote_ does not have the same meaning as in French.

What about keeping `Election` for the whole process and `Ballot` for `Vote`?

authentication/models/user.py
Piwakk: Indeed!

PR #104, opened 2019-09-20
17wf/rewrite associations urls

Piwakk:


api/urls.py
```diff
@@ -6,17 +6,17 @@
 import chat.urls
 import forum.urls
 import polls.urls
-import repartitions.urls
+#import repartitions.urls
```
abocquet: could be removed

PR #105, opened 2019-09-21
Remove News and add a type to Page.

Piwakk:


associations/models/page.py
```diff
@@ -1,12 +1,30 @@
...
+
+    NEWS = 'NEWS'
+    """A Page which will be displayed in the news feed of the association."""
+
+    STATIC = 'STATIC'
+    """A Page which can be accessed from the association main page."""
+
+    PAGE_TYPES = (
+        (NEWS, 'News page'),
+        (STATIC, 'Static page')
```
abocquet: Maybe not useful to have a "display name" since it will be handled in the front-end. Might as well define 
```
PAGE_TYPES = {
     "NEWS" = "NEWS",
     "STATIC" = "STATIC"
}
```

associations/views/page.py
```diff
@@ -9,11 +9,10 @@ class PageViewSet(AssociationNestedViewSet):
...
+        # Send the current user to the serializer so they can be added to the authors.
+        serializer.save(author=self.request.user)
 
     def perform_update(self, serializer):
-        serializer.save(association=Association.objects.get(pk=self.kwargs['association_pk']))
-
-    def perform_create(self, serializer):
-        serializer.save(association=Association.objects.get(pk=self.kwargs['association_pk']))
+        # Send the current user to the serializer so they can be added to the authors.
+        serializer.save(author=self.request.user)
```
abocquet: Does it work to says that `author` is a specific user since it's a many to many relationship ? Should be make something like `serializer.instance.authors.add(self.request.user)` ?

associations/models/page.py
Piwakk: Well, they might be useful in the admin pane at least. You never know, and I thought it does no harm to keep them here.

associations/views/page.py
Piwakk: The `serializer.save` method was overrided:
```python
    def save(self, author, **kwargs):
        """Override the save method to add `author` to the authors field."""
        instance = super(PageSerializer, self).save(**kwargs)
        instance.authors.add(author)
        return instance
```

PR #106, opened 2019-09-23
17wf/testsAssociation

Piwakk:


associations/permissions/marketplace.py
```diff
@@ -1,173 +1,140 @@
...
+                return product_query[0].marketplace.enabled
+            else:
+                raise NotFound('The requested product does not exist.')
+
+        return True
+
+    def has_object_permission(self, request, view, transaction):
+        # We have to put it a second time here, because when retrieving, both has_permission and has_object_permission
+        # will be called.
+        if request.method in SAFE_METHODS:
```
abocquet: Not everybody should be able to see the transactions of everyone, only themselves and the admins

associations/permissions/marketplace.py
```diff
@@ -1,173 +1,140 @@
...
+                return market.enabled and role and role.marketplace
+            else:
+                raise NotFound('The requested marketplace does not exist.')
+
+        return True
+
+    def has_object_permission(self, request, view, funding):
+        # We have to put it a second time here, because when retrieving, both has_permission and has_object_permission
+        # will be called.
+        if request.method in SAFE_METHODS:
```
abocquet: Same comment here for the same reasons

associations/serializers/association.py
```diff
@@ -1,110 +1,108 @@
...
+        res = super(RoleSerializer, self).to_representation(instance)
+
+        for permission_name in Role.PERMISSION_NAMES:
+            res[permission_name] = getattr(instance, permission_name)
+
+        return res
+
+    class Meta:
+        model = Role
+        fields = '__all__'
```
abocquet: It's possible to change the `association` with this serializer. Perhaps it's better to change this to removing `to_representation` and declaring all fields as `read_only_field`(https://stackoverflow.com/questions/38319856/django-rest-framework-read-only-model-serializer)

associations/permissions/marketplace.py
Piwakk: Actually, this is enforced in the view by `get_queryset`.
But maybe it is a bit unclear (and it returns a 404 instead if a 403, which I personnally don't mind).
Do you think I should add a comment? Or explictly forbid in the permission, even if it's redundant?

Piwakk:
It is indeed.
Problem is, the fields in `to_representation` are actually properties, and DRF doesn't seem to consider them as regular fields...
But you are right, this Serializer should definitely be read-only, as one may be using it for writing data without paying attention (actually happened to me...) Maybe we could override `save` to raise an exception if it is used?

associations/serializers/association.py
Piwakk: It is indeed.
Problem is, the fields in to_representation are actually properties, and DRF doesn't seem to consider them as regular fields...
But you are right, this Serializer should definitely be read-only, as one may be using it for writing data without paying attention (actually happened to me...) Maybe we could override save to raise an exception if it is used?

associations/permissions/marketplace.py
abocquet: Redundancy is not a bad thing when it comes to such an important security matter

abocquet: Also if for some reason the view came to be changed and did not enforce this, the information would become public

Piwakk: OK let's do this

PR #109, opened 2019-09-25
17wf/polls

Piwakk:


polls/models.py
```diff
@@ -44,26 +47,39 @@ class Poll(models.Model):
...
         """
-        Return True if the poll is has been published in the past, that is publication_date_time <= now(),
-        and its state is ACCEPTED.
+        Return True if the poll is has been published in the past, that is publication_date <= now(), and its state is
+        ACCEPTED.
         """
         td = date.today()
         if self.publication_date is None or self.state != 'ACCEPTED':
             return False
         return self.publication_date <= td
```
abocquet: No need to declare a tmp variable that will be read once, not modified and short to write

polls/permissions.py
```diff
@@ -0,0 +1,70 @@
...
+
+            return (request.user.is_staff or  # Administrators can always retrieve.
+                    request.user == poll.user or  # Authors can always retrieve.
+                    poll.has_been_published)  # Published polls are public.
+        elif request.method in ('PATCH',):
+            self.message = 'You are not allowed to update this poll.'
+
+            return (request.user.is_staff or  # Administrators can always update.
+                    (request.user == poll.user and
+                     not poll.state == 'ACCEPTED'))  # Authors can only update if the poll is not accepted.
```
abocquet: Are `ACCEPTED` and `PENDING` the two only states of a poll ? If not, this permission might be leaky

Piwakk: Three states: `REVIEWING` (not `PENDING`), `ACCEPTED` and `REJECTED`.
So yes, let's forbid the update if the state is `REJECTED`.

Piwakk: Three states: `REVIEWING` (not `PENDING`), `ACCEPTED` and `REJECTED`.
So yes, let's forbid the update if the state is `REJECTED`.

PR #111, opened 2019-09-25
Abocquet/tags

abocquet:


tags/models.py
```diff
@@ -0,0 +1,70 @@
...
+    Folder,
+    Loan,
+    Association,
+)
+from forum.models import Theme, Topic, MessageForum
+
+
+class Namespace(models.Model):
+    class Meta:
+        unique_together = (("name", "scope", "scoped_to"),)
```
Piwakk: See https://docs.djangoproject.com/en/2.2/ref/models/options/#unique-together, `unique_together` can be reduced to a single tuple.

tags/models.py
```diff
@@ -0,0 +1,70 @@
...
+    )
+
+    value = models.CharField(max_length=50)
+    url = models.CharField(max_length=255, null=True, blank=True)
+
+    class Meta:
+        unique_together = (("value", "namespace"),)
+
+
+# Dynamically adds the needed fiields to the tag
```
Piwakk: Typo in the comment

tags/permissions.py
```diff
@@ -0,0 +1,126 @@
...
+def user_can_link_tag_to(user: User, tag: Tag, instance):
+    parent = get_parent_object(instance)
+    if (
+        not isinstance(parent, Namespace.SCOPES[tag.namespace.scope])
+        or parent.id != tag.namespace.scoped_to
+    ):
+        return False
+
+    if isinstance(parent, Association):
+        role = _get_role_for_user(user, parent)
```
Piwakk: Sorry, forgot to tell you: the `User` model now has a method (`User.get_role`) specifically for this. You may want to use it instead of another function, which I think is now useless

tags/permissions.py
```diff
@@ -0,0 +1,126 @@
...
+            scope = request.data.get("scope")
+            scoped_to = request.data.get("scoped_to")
+
+            if scope != "global" and (scope is None or scoped_to is None):
+                raise SuspiciousOperation()
+
+            if scope == "global":
+                return request.user.is_admin
+
+            instance = Tag.LINKS[scope].objects.get(pk=scoped_to)
```
Piwakk: What happens if `scoped_to` does not exist in the database ? Django will raise `ObjectDoesNotExist`; maybe catch it to raise another more meaningful exception? (For now I have just started the review and I see this file first, sorry if this is handled elsewhere).

tags/permissions.py
```diff
@@ -0,0 +1,126 @@
...
+            return can_manage_tags_for(request.user, instance)
+
+        return True
+
+    def has_object_permission(self, request, view, namespace):
+        if namespace.scope == "global":
+            return request.user.is_admin
+
+        instance = Namespace.SCOPES[namespace.scope].objects.get(pk=namespace.scoped_to)
```
Piwakk: Same here

tags/permissions.py
```diff
@@ -0,0 +1,126 @@
...
+        return can_manage_tags_for(request.user, instance)
+
+
+class ManageTagPermission(permissions.BasePermission):
+    def has_permission(self, request, view):
+        if request.method == "POST":
+            namespace = Namespace.objects.get(pk=request.data.get("namespace"))
```
Piwakk: Same here

Piwakk: Well I think you understood now, there are two or three more afterwards, I don't signal them ;)

tags/models.py
```diff
@@ -0,0 +1,70 @@
...
+
+class Namespace(models.Model):
+    class Meta:
+        unique_together = (("name", "scope", "scoped_to"),)
+
+    SCOPES = {"association": Association, "forum_theme": Theme, "global": None}
+
+    name = models.CharField(max_length=50)
+
+    scope = models.CharField(max_length=50)  # The application it targets
```
Piwakk: Maybe use `choices=SCOPES.keys()` here

tags/serializers.py
```diff
@@ -0,0 +1,70 @@
...
+
+class NamespaceSerializer(serializers.ModelSerializer):
+    class Meta:
+        model = Namespace
+        fields = ("id", "name", "scope", "scoped_to")
+
+    def validate_scope(self, value):
+        if value not in Namespace.SCOPES.keys():
+            raise serializers.ValidationError("The scope isn't valid")
+        return value
```
Piwakk: See my comment on the `Namespace` model, DRF will do it by itself if the model knows that `scope` must be a key of `Namespace.SCOPES`.

tags/serializers.py
```diff
@@ -0,0 +1,70 @@
...
+
+    def validate(self, data):
+        if len(data) == 0:
+            return data
+
+        if bool(Namespace.SCOPES.get(data.get("scope", None), False)) != bool(
+            data.get("scoped_to", False)
+        ):  # logical XOR
```
Piwakk: Weird formatting here (also in the next tests)

tags/serializers.py
```diff
@@ -0,0 +1,70 @@
...
+            raise serializers.ValidationError(
+                "Must define scope and scoped_to together"
+            )
+
+        if Namespace.SCOPES.get(data.get("scope")) and not Namespace.SCOPES.get(
+            data.get("scope")
+        ).objects.get(pk=data.get("scoped_to")):
```
Piwakk: IMHO, one thing which could help to read this whole function would be to define the local variables `scope = data.get("scope", None)` and `scoped_to = data.get(…)`

tags/serializers.py
```diff
@@ -0,0 +1,70 @@
...
+            raise serializers.ValidationError("Not scoped_to and existing object")
```
Piwakk: Typo in Exception message?

tags/serializers.py
```diff
@@ -0,0 +1,70 @@
...
+
+    def to_representation(self, instance):
+        response = super(TagSerializer, self).to_representation(instance)
+        response["namespace"] = NamespaceSerializer().to_representation(
+            instance.namespace
+        )
+        return response
+
+    def update(self, instance, validated_data):
+        raise NotImplementedError("Cannot update a tag from the REST API")
```
Piwakk: Maybe explain why here in a comment?

tags/models.py
```diff
@@ -0,0 +1,70 @@
...
+        "topic": Topic,
+        "message_forum": MessageForum,
+    }
+
+    namespace = models.ForeignKey(
+        Namespace, related_name="tags", on_delete=models.CASCADE
+    )
+
+    value = models.CharField(max_length=50)
+    url = models.CharField(max_length=255, null=True, blank=True)
```
Piwakk: What is the purpose of this field?

tags/models.py
```diff
@@ -0,0 +1,70 @@
...
+
+    class Meta:
+        unique_together = (("value", "namespace"),)
+
+
+# Dynamically adds the needed fiields to the tag
+for (link_name, link_class) in Tag.LINKS.items():
+    field = models.ManyToManyField(link_class, related_name="tags")
+    field.contribute_to_class(Tag, link_name)
```
Piwakk: Can you explain `contribute_to_class`? I feel like it is not a common pattern in Django and I can't find anything in the doc about it

tags/tests/test_link_tag.py
```diff
@@ -0,0 +1,96 @@
...
+
+    def test_add_global_tag(self):
+        self.login("17simple")
+        res = self.post("/tags/link/product/2/tag/1/")
+        self.assertStatusCode(res, 201)
+
+        res = self.get("/associations/products/2/")
+        self.assertStatusCode(res, 200)
+        self.assertEqual(
+            json.loads(res.content).get("tags"),
```
Piwakk: I think it applies to your case, usually `res.data` contains the deserialized Python objects, so you don't have to use `json.loads`

tags/tests/test_manage_tag.py
```diff
@@ -0,0 +1,126 @@
...
+
+        self.login("17admin")
+        res = self.delete("/tags/tag/{}/".format(tag["id"]))
+        self.assertStatusCode(res, 403)
+
+        self.login("17admin_pdm")
+        res = self.delete("/tags/tag/{}/".format(tag["id"]))
+        self.assertStatusCode(res, 204)
+
+    def test_create_same_tage_twice(self):
```
Piwakk: typo here

tags/urls.py
```diff
@@ -0,0 +1,19 @@
+from django.urls import path
+from rest_framework_bulk.routes import BulkRouter
+
+from tags.views import NamespaceViewSet, TagViewSet, TagLinkView, get_tags_for_scope
+
+urlpatterns = [
+    path(
+        "link/<model>/<instance_pk>/tag/<tag_pk>/",
```
Piwakk: You can specify the format of the capturing `<…>` (https://docs.djangoproject.com/en/2.2/topics/http/urls/#path-converters)

tags/models.py
Piwakk: See https://docs.djangoproject.com/en/2.2/ref/models/options/#unique-together, `unique_together` can be reduced to a single tuple.

tags/views.py
```diff
@@ -0,0 +1,115 @@
...
+    queryset = Namespace.objects.all()
+    serializer_class = NamespaceSerializer
+    filter_backends = (DjangoFilterBackend,)
+    filterset_fields = ("scope", "scoped_to")
+    permission_classes = (NamespacePermission,)
+
+    def create(self, request, *args, **kwargs):
+        # Required since we use "unique_together"
+        if "scoped_to" not in request.data:
+            request.data["scoped_to"] = ""
```
Piwakk: Er… how can a Namespace be `scoped_to` nothing actually?

tags/views.py
```diff
@@ -0,0 +1,115 @@
...
+        return queryset
+
+    def create(self, request, *args, **kwargs):
+        if "namespace" in request.data and "value" in request.data:
+            try:
+                tag = Tag.objects.get(
+                    namespace=request.data.get("namespace"),
+                    value=request.data.get("value"),
+                )
+                return JsonResponse(TagSerializer().to_representation(tag), status=201)
```
Piwakk: I think it's better to use the DRF object `rest_framework.response.Response`, at least we are sure it will integrate well with the DRF framework (in particular the browsable API)
I saw you imported `django.HttpResponse`, same for this one

tags/models.py
abocquet: It was here to determine where to redirect the user if they click on the link. Might be created dynamically though

tags/models.py
abocquet: It's like `setattr(Tag, link_name)` but it makes Django ORM take the field into account

tags/views.py
abocquet: Yes, if the scope is global

tags/permissions.py
abocquet: I see why it's bothering but I don't know how to come with a solution more elegant that `try ... except` all the time, which would make the code less clear IMO

tags/serializers.py
abocquet: Black formats as it, let's trust it

tags/models.py
abocquet: Added some comments in the code

tags/serializers.py
abocquet: I'm not sure I quite get it

Piwakk: I mean, if you define the `scope` field of the `Namespace` model like this:
`scope = models.CharField(…, choices=SCOPES.keys())`
then DRF will enforce your `validate_scope` function by itself

associations/views/association.py
```diff
@@ -65,3 +65,13 @@ def get_serializer_class(self):
...
+
+def upload_association_logo(request, association_pk):
+    if request.method == 'POST' and request.FILES['myfile']:
+        myfile = request.FILES['myfile']
+        fs = FileSystemStorage()
+        filename = fs.save(myfile.name, myfile)
+        uploaded_file_url = fs.url(filename)
+        return render(request, 'core/simple_upload.html', {
+            'uploaded_file_url': uploaded_file_url
+        })
```
Piwakk: Er, it's weird to see `upload_association_logo` here, but I guess it's a remnant from the other branch

abocquet: Hum true 😅

PR #112, opened 2019-09-26
Add association image

abocquet:


associations/urls.py
```diff
@@ -36,6 +36,7 @@
 router.register(r'transactions', TransactionViewSet)
 router.register(r'funding', FundingViewSet)
 urlpatterns += [
+    path("associations/<association_pk>/image", set_association_image),
```
Piwakk: Is the omission of the trailing `/` intentional?

associations/views/association.py
```diff
@@ -56,7 +64,21 @@ class AssociationViewSet(viewsets.ModelViewSet):
...
+    association = Association.objects.get(pk=association_pk)
+    role = request.user.get_role(association)
+    if not (role and role.administration):
+        raise PermissionDenied()
+
+    serializer = AssociationImageSerializer(data=request)
+    serializer.is_valid(raise_exception=True)
+    serializer.update(association, serializer.validated_data)
+
+    return Response(status=200)
```
Piwakk: Every time I test a create function, the default response of DRF is a 201, so I believe it should be the same here

associations/urls.py
abocquet: nope

associations/views/association.py
abocquet: this is not a creation, but an update. Maybe the code should not be 200, but I beleive it's not 201 either

Piwakk: Actually, maybe the http verb to use is `PUT`.
Also see the MDN for the response codes (yeah, there are several cases…) https://developer.mozilla.org/fr/docs/Web/HTTP/M%C3%A9thode/PUT#R%C3%A9ponses

associations/urls.py
```diff
@@ -56,6 +57,7 @@
 router.register(r"transactions", TransactionViewSet)
 router.register(r"funding", FundingViewSet)
 urlpatterns += [
+    path("associations/<association_pk>/image", set_association_logo),
     path("marketplace/balance/", BalanceView.as_view(), name="balance-list"),
```
Piwakk: No trailing / ?
Sorry, I forgot what you said last time about it

PR #113, opened 2019-09-27
[Tag] secret management (example with events)

abocquet:


tags/filter_mixins.py
```diff
@@ -0,0 +1,49 @@
...
+            "'%s' should either include a `queryset` attribute, "
+            "or override the `get_queryset()` method." % self.__class__.__name__
+        )
+
+        queryset = self.queryset
+
+        if isinstance(queryset, QuerySet):
+            if (
+                queryset.model in Tag.LINKS.values()
+                and self.request.user.is_in_first_year
```
Piwakk: Rather user `User.show`, which takes `settings.SHOW_TO_FIRST_YEAR_STUDENTS` into account

tags/middleware.py
```diff
@@ -0,0 +1,49 @@
+from django.db.models import Q
+from tags.models import Tag
+
+
+class InjectTagFilteringMiddleware:
```
Piwakk: Explain what it does

tags/middleware.py
```diff
@@ -0,0 +1,49 @@
...
+    hiding_conditions = {
+        "association": {"tags__is_hidden": True},
```
Piwakk: Why using a `dict` and not simply `Q(tags__is_hidden=True)`?

tags/middleware.py
```diff
@@ -0,0 +1,49 @@
...
+                    Q(**hiding_condition)
+                    if not isinstance(hiding_condition, Q)
+                    else hiding_condition
+                )
+
+                model.objects = model.objects.filter(
+                    Q(tags__is_hidden=True) | hiding_condition
+                )
+
+            Tag.objects = Tag.objects.filter(is_hidden=True)
```
Piwakk: I'm not sure I understand. Here, you are only keeping the database objects which should be hidden, right?

Piwakk: But this code is executed when the user should *not* be shown the things.
So should you not use `exclude` instead of `filter`?

tags/models.py
```diff
@@ -94,3 +93,6 @@ def get_linked_instance(scope, scoped_to):
 for (link_name, link_class) in Tag.LINKS.items():
     field = models.ManyToManyField(link_class, related_name="tags")
     field.contribute_to_class(Tag, link_name)
+
+    field = models.ManyToManyField(link_class, related_name="inherited_tags")
+    field.contribute_to_class(Tag, link_name + "_by_inheritance")
```
Piwakk: Can you explain what is it?

tags/middleware.py
abocquet: It's a middle ware that injects a tag filter 😇

PR #117, opened 2019-10-07
SSO Integration

Piwakk:


authentication/authentication.py
```diff
@@ -1,76 +1,74 @@
...
+            raise AuthenticationFailed("Custom header against CSRF attacks is not set.")
 
-        raw_token = request.COOKIES.get(api_settings.ACCESS_TOKEN_COOKIE_NAME)
+        # Get the cookie.
+        raw_token = request.COOKIES.get(
+            settings.JWT_AUTH_SETTINGS["ACCESS_TOKEN_COOKIE_NAME"]
+        )
         if raw_token is None:
-            raise AuthenticationFailed("Authorization cookie not set")
-        validated_token = self.get_validated_token(raw_token)
```
abocquet: no need for token validation ?

authentication/keys/public.py
```diff
@@ -0,0 +1,9 @@
+PUBLIC_KEY = b"""-----BEGIN PUBLIC KEY-----
+MIIBIjANBgkqhkiG9w0BAQEFAAOCAQ8AMIIBCgKCAQEAoWPv5B06KeQPru1unPs1
+k/iAME1nIEsoWJIkmaP5broZcUpztOpmRAf8+RCk5oAXQrJ87SoIs8UG4fKCduNa
+FXsKvv1zn6z+S92ScpHOYmuxO9mlL1np9NbhRdiSDwIomdpXNZ+ttF+Ll0nO87wU
+aKvTdwLhLswSLEjt2YwmKOhgpSPe2mLkk5Me1v7iAfj4ermisaV/WLu7QbelboZO
+Yp3yhOkPVKhvJU4fBkx/cnkcvLMo8dx5sM3TQonZB5oiv1w4YWz7nBjbmH3bv+9u
+AiLpr9K2ueieZbM1R3f7vhS44pNUI5LYRYA4NHEv1O+n5bE8BuvnVMG5qGP6fB8J
+uwIDAQAB
+-----END PUBLIC KEY-----"""
```
abocquet: Let's not put SSH keys directly in the code but add a CLI tool to generate this file when needed. That way we'll ensure dev and prod have different keys

authentication/views/authentication.py
```diff
@@ -0,0 +1,73 @@
...
+import jwt
+from rest_framework import status, views
+from rest_framework.exceptions import AuthenticationFailed
+from rest_framework.response import Response
+
+from authentication.token import decode_token
+
+
+class LoginView(views.APIView):
+    """This view takes a JWT as a GET parameter, verifies it and sets it as a cookie."""
```
abocquet: There should be some kind of check such as a state parameter to make sure the JWT comes from the SSO server. Otherwise this could be a vector for an attack
https://auth0.com/docs/protocols/oauth2/oauth-state

polls/tests.py
```diff
@@ -1,10 +1,10 @@
 from datetime import date, datetime, timezone
 
-from backend.tests_utils import BaseTestCase
+from backend.tests_utils import WeakAuthenticationBaseTestCase
 from polls.models import Poll
 
 
-class PollTestCase(BaseTestCase):
+class PollTestCase(WeakAuthenticationBaseTestCase):
```
abocquet: Is there a way to avoid to change all `BaseTestCase` to `WeakAuthenticationBaseTestCase` ?

authentication/keys/public.py
Piwakk: This key depends on the private key which is on the SSO, thus generated somewhere else
I don't know how to do this cleanly

Piwakk: I mean, for me, it's clean enough, like, do we have another option than manually transferring the public key?

authentication/views/authentication.py
Piwakk: There is an `iss` claim + the RSA signing, isn't that enough?

polls/tests.py
Piwakk: I could have left it as `BaseTestCase`, but I thought changing the name was important as we are disabling a major security feature here, and we don't want to use `WeakAuthenticationBaseTestCase` for testing, well, the authentication security

authentication/authentication.py
Piwakk: It is done when calling `decode_token`

authentication/views/authentication.py
Piwakk: (which are all checked for in `decode_token`)

abocquet: The state will prevent CSRF attacks, it has nothing to do with encryption. Currently someone could send the user a link and set arbitrary cookies in their browser

polls/tests.py
abocquet: Is there a way not to remove theses security feature ?

authentication/keys/public.py
abocquet: Maybe not, but in that case just put a string that is explicitly not a key so it will fail if someone tries to use the code as it

Piwakk: Sorry, what do you mean?

authentication/views/authentication.py
Piwakk: Correct me if I'm wrong, but I read the code once again, and I see this logic:
if the GET parameter is not correctly decoded by `decode_token`, raise an Exception (and thus don't set the cookie)
So I believe an arbitrary string (and actually anything that is not a valid token) will not be set as a cookie

polls/tests.py
Piwakk: Another approach would be to replace the PUBLIC_KEY settings indeed
I'll do that

authentication/keys/public.py
abocquet: In summary : don't put public or private keys in the code. If you need to make then explicitly wrong ("dummy public key" for instance)

authentication/tests/test_authentication.py
```diff
@@ -1,23 +1,107 @@
...
+
+    def assertIsNotAuthenticated(self):
+        res = self.get("/auth/check/")
+        self.assertStatusCode(res, 401)
+
+    #########
+    # LOGIN #
+    #########
+
+    def test_if_no_token_then_cannot_login(self):
```
abocquet: `cannot_login_without_token`

authentication/tests/test_authentication.py
```diff
@@ -1,23 +1,107 @@
...
+        res = self.get("/auth/login/")
+        self.assertStatusCodeIn(res, [401, 403])
+        self.assertEqual(
+            res.data["detail"],
+            "Provide a JWT as the `access` GET parameter.",
+            msg=res.data["detail"],
+        )
+
+    def test_if_invalid_signature_token_then_cannot_login(self):
```
abocquet: `cannot_login_with_invalid_signature`

authentication/tests/test_authentication.py
```diff
@@ -1,23 +1,107 @@
...
+        res = self.get("/auth/check/")
+        self.assertStatusCode(res, 401)
+
+    #########
+    # LOGIN #
+    #########
+
+    def test_if_no_token_then_cannot_login(self):
+        res = self.get("/auth/login/")
+        self.assertStatusCodeIn(res, [401, 403])
```
abocquet: Why don't we know the status code ?

authentication/tests/test_authentication.py
```diff
@@ -1,23 +1,107 @@
...
+        self.assertEqual(
+            res.data["detail"],
+            "Provide a JWT as the `access` GET parameter.",
+            msg=res.data["detail"],
+        )
+
+    def test_if_invalid_signature_token_then_cannot_login(self):
+        res = self.login(self.tokens["INVALID_SIGNATURE_TOKEN"])
+        self.assertStatusCodeIn(res, [401, 403])
```
abocquet: Same

authentication/tests/test_authentication.py
```diff
@@ -1,23 +1,107 @@
...
+
+    # These tokens were generated with the private key corresponding to `authentication.keys.public.PUBLIC_KEY`. This
+    # private key should be in the `sso` project.
+    # One can use the script in the `sso_server` project to generate those tokens, and https://jwt.io/ to debug them.
+    # The audience of these tokens is `portail` (it was of course edited in INVALID_AUDIENCE_TOKEN).
+    # Except for `VALID_17UNKNOWN_TOKEN`, the user of these tokens is `17simple`.
+
+    # tokens generated with `sso_server.tests.generate_fake_tokens("portail", "17simple")` in the `sso` project.
+    tokens = {
+        "INVALID_SIGNATURE_TOKEN": "eyJ0eXAiOiJKV1QiLCJhbGciOiJSUzI1NiJ9.eyJleHAiOjQ3MjYwMjE5ODQuNzk3MzMzLCJpc3MiOiJzc29fc2VydmVyIiwiYXVkIjoicG9ydGFpbCIsImp0aSI6ImJjOTM4Yzc1NjE4NTRiZTU4MTc4YzZiNzVmNWZjNWNkIiwidXNlciI6IjE3c2ltcGxlIn0.Xu4m9AA3R_y_7RN_7v3rrNq9NQ6f0aeOCTbvQp-HZBR6PhQamhSyAj9vZD05oTLl-ManzDvIXjg8QbKZh3-N0HoO5yWUHtM0CKWJUQ07NptH_VldOtvbc8Fb4Lhzr03ScxtQslaUc6DXraWd3GurDCUGRzw9NT7_VybycofwucjvQjE1gFbjqN7TnxAdb8Tye5w-89ge-1FGZvsqgZyvupScdEe7B15QUe2UTm-RsDDV-RG91B6TBoTduFpTzp3oqzrkwONurKjK929Z-Ndq6AHOP17Cj_SL4R6rxlpVsQ50j0LMODhxwBfIHnQsARiLdwGcThl6LgZLYLfciAAAAA",
```
abocquet: Maybe use a function to encode the payload here so we know what's in the tokens

authentication/views/authentication.py
```diff
@@ -0,0 +1,73 @@
...
+
+    authentication_classes = ()
+    permission_classes = ()
+
+    def get(self, request, format=None, *args, **kwargs):
+        if settings.JWT_AUTH_SETTINGS["GET_PARAMETER"] in request.GET:
+            # Retrieve the token.
+            token = request.GET.get(settings.JWT_AUTH_SETTINGS["GET_PARAMETER"])
```
abocquet: You can use a `[]` syntax instead of `get` since you know the keys is in `request.GET`. Another way could be 
```suggestion
            token = request.GET.get(settings.JWT_AUTH_SETTINGS["GET_PARAMETER"], None)
            if token:                
```

backend/fake_private_key.py
```diff
@@ -0,0 +1,29 @@
+"""Used by the tests to generate fake JWT."""
+
+FAKE_PRIVATE_KEY = b"""-----BEGIN RSA PRIVATE KEY-----
```
abocquet: Do we need the `.pem` file ? If yes can we load `backend/fake_private_key.pem` into this file instead of needing to copy paste them ?

backend/settings.py
```diff
@@ -16,26 +16,26 @@
...
+env = environ.Env()
+environ.Env.read_env()
+
 BASE_DIR = os.path.dirname(os.path.dirname(os.path.abspath(__file__)))
 
 # Quick-start development settings - unsuitable for production
 # See https://docs.djangoproject.com/en/2.0/howto/deployment/checklist/
 
 # SECURITY WARNING: keep the secret key used in production secret!
-SECRET_KEY = os.environ.get(
```
abocquet: What is the advantage of `environ` over `os.environ` ?

backend/settings.py
```diff
@@ -122,11 +112,11 @@
 DATABASES = {
     "default": {
         "ENGINE": "django.db.backends.postgresql",
-        "NAME": os.environ.get("DATABASE_NAME", "portail"),
-        "USER": os.environ.get("DATABASE_USER", "root"),
-        "PASSWORD": os.environ.get("DATABASE_PASSWORD", "password"),
-        "HOST": os.environ.get("DATABASE_HOST", "localhost"),
-        "PORT": os.environ.get("DATABASE_PORT", "5432"),
+        "NAME": env.str("DATABASE_NAME"),
```
abocquet: Since we don't put default values, maybe create a `.env.dist` file with dummy values so we know what fields we need to fill

backend/tests_utils.py
```diff
@@ -32,92 +41,75 @@ def assertStatusCodeIn(self, res, status_codes, user_msg=""):
...
+            algorithm=settings.JWT_AUTH_SETTINGS["ALGORITHM"],
+        ).decode(
+            "utf-8"
+        )  # jwt.encode returns a bytes object, it thus has to be decoded to a str.
 
-    def patch(self, url, data=None, format="json", content_type="application/json"):
-        return self.client.patch(
-            self.api_base + url, data, format=format, content_type=content_type
-        )
+    def login(self, username, password="password"):
```
abocquet: We may not need the `password` keyword now

backend/fake_private_key.py
Piwakk: We don't need the PEM file. Let's just use the PY file.

authentication/tests/test_authentication.py
Piwakk: This is quantum computing
(jkd, it should be a 403)

authentication/tests/test_authentication.py
Piwakk: Well, actually, it is more or less well explained above.

These test tokens could be generated in the `Portail`, but this requires to give the project a pair of public / private key, etc.
For now, it works well like this. I will do it but in another PR before production.

authentication/tests/test_authentication.py
Piwakk: I mean, a 401

backend/settings.py
Piwakk: You mean, for `SECRET_KEY`?
`environ` uses a `.env` file
I don't know if you can achieve the same with built-in `os.environ`, I don't think so

authentication/tests/test_authentication.py
abocquet: Please at least add in comment the unencoded tokens

backend/settings.py
abocquet: Ok 👍

backend/settings.py
Piwakk: Good idea!

backend/tests_utils.py
Piwakk: OK, I will just check if it doesn't break any test

authentication/tests/test_authentication.py
Piwakk: For some reason, raising `AuthenticationFailed` returns a 403.
Whatever.

backend/tests_utils.py
Piwakk: It was indeed useless

authentication/tests/test_authentication.py
abocquet: Then put `403` and explain it should be a `401`

PR #118, opened 2019-10-27
Repartitions

abocquet:


repartitions/algorithm.py
```diff
@@ -0,0 +1,200 @@
...
+    s = 0
+    for i in range(len(places)):
+        if s <= index < s + places[i]:
+            return i
+        s += places[i]
+
+    if index == s:
+        return len(places) - 1
+
+    raise Exception("should not happend")
```
Piwakk: happen

Piwakk: Better to raise a more specific Exception by the way

repartitions/algorithm.py
```diff
@@ -0,0 +1,200 @@
...
+
+
+def generate_line(
+    uc: UserCampaign, propositions: List[Proposition], places: List[int]
+) -> List[float]:
+    if any(map(lambda x: x < 0, places)):
+        raise Exception("places {} is not valid", places)
```
Piwakk: It seems like a `ValueError`

repartitions/algorithm.py
```diff
@@ -0,0 +1,200 @@
...
+    return res
+
+
+def make_repartition_for_category(
+    category: Category, propositions: List[Proposition], already_used: List[int]
+) -> List[List[UserCampaign]]:
+    raw_users_campaign = category.users_campaign.all()
+
+    if len(raw_users_campaign) == 0:
+        raise Exception("No users in given group")
```
Piwakk: ValueError

repartitions/algorithm.py
```diff
@@ -0,0 +1,200 @@
...
+            repartition_cardinal_diff = 0  # avoid being stuck in an infinite loop
+
+    print_repartition = []
+    for i in repartition:
+        print_repartition.append(len(i))
+
+    return repartition
+
+
+# Checks the repartition is always even
```
Piwakk: Docstring

repartitions/algorithm.py
```diff
@@ -0,0 +1,200 @@
+from typing import List
```
Piwakk: So, my comments on these files are only about the typos, and this is because I just cannot understand the code.

I completely trust you (and the tests) about the algorithm, but if you want to give your own code a chance to not being dropped by the future maintainers of this project, please add comprehensive docstrings (which also explain the variables) and some comments; do not hesitate to be talkative.

At a first glance, we cannot even understand what is the main function, let alone the various `i`, `j`, `s`… (maybe there is not a better name for them, in that case explain thoroughly what the loop does before).

Thanks

repartitions/algorithm.py
```diff
@@ -0,0 +1,200 @@
...
+    already_used = [0] * len(propositions)
+    store = {}
+
+    for i in range(len(categories)):
+        category = categories[i]
+
+        if category.users_campaign.count() == 0:
+            continue
+
+        # Check the repartition is feasible
```
Piwakk: Make it a function then?

repartitions/models.py
```diff
@@ -0,0 +1,68 @@
...
+from authentication.models import User
+
+
+class Campaign(models.Model):
+    STATUS = ("CLOSED", "OPEN", "RESULTS")
+
+    id = models.AutoField(primary_key=True)
+
+    name = models.CharField(max_length=200)
+    status = models.CharField(max_length=200, default="CLOSED")
```
Piwakk: choices?

repartitions/models.py
```diff
@@ -0,0 +1,68 @@
...
+    )
+
+
+class Proposition(models.Model):
+    id = models.AutoField(primary_key=True)
+
+    campaign = models.ForeignKey(
+        Campaign, on_delete=models.CASCADE, related_name="propositions"
+    )
+    name = models.CharField(max_length=200, blank=True, null=True, default=None)
```
Piwakk: Is it meaningful to have a blank or null proposition?

repartitions/models.py
```diff
@@ -0,0 +1,68 @@
...
+class Campaign(models.Model):
+    STATUS = ("CLOSED", "OPEN", "RESULTS")
+
+    id = models.AutoField(primary_key=True)
+
+    name = models.CharField(max_length=200)
+    status = models.CharField(max_length=200, default="CLOSED")
+    even = models.BooleanField(
+        default=True
+    )  # should  all the categories be evenly split between groups
```
Piwakk: This comment should go to `help_text`, at least it has a chance to be displayed somewhere in Django. Here, it will not even be displayed by the IDE's

repartitions/models.py
```diff
@@ -0,0 +1,68 @@
+from django.db import models
```
Piwakk: _Please_ add docstrings to explain what are these models.
What is the difference between a `Campaign`, a `UserCampaign`? Between a `Category` or a `Group` or a `Proposition`? There are six models here and their relationship is not linear, it is very hard to understand.

repartitions/permissions.py
```diff
@@ -0,0 +1,58 @@
...
+def user_in_campaign(user: User, campaign: Campaign) -> bool:
+    for category in campaign.categories.all():
+        if user in [uc.user for uc in category.users_campaign.all()]:
+            return True
+
+    return False
+
+
+class CanManageCampaign(BasePermission):
+    message = "Editing association is not allowed."
```
Piwakk: Campaign?

repartitions/permissions.py
```diff
@@ -0,0 +1,58 @@
...
+            return request.user == obj.user
+
+        return False
+
+
+class UserCampaignPermission(BasePermission):
+    def has_permission(self, request, view):
+        if request.method in SAFE_METHODS:
+            return True
+        return request.user.is_admin
```
Piwakk: I think you can afford the one-liner `return request.user.is_admin or request.method in SAFE_METHODS` here

repartitions/views.py
```diff
@@ -0,0 +1,196 @@
...
+
+    def perform_create(self, serializer):
+        campaign = Campaign.objects.get(pk=self.kwargs["parent_lookup_campaign"])
+        serializer.save(campaign=campaign)
+
+    def get_object(self):
+        """
+        Returns the object the view is displaying.
+
+        You may want to override this if you need to provide non-standard
```
Piwakk: So I guess this was copy-pasted from the library source, what is the addition?

repartitions/views.py
```diff
@@ -0,0 +1,196 @@
+import json
+
+from django.http import JsonResponse, HttpResponseBadRequest
```
Piwakk: Why do you use `JsonResponse` instead of DRF's `Response`?
I understand for `HttpResponseBadRequest`, because DRF's equivalent is `ValidationError` which does not convey the same meaning

repartitions/tests/munkres.py
```diff
@@ -0,0 +1,302 @@
...
+import numpy as np
+from nose.tools import raises
+
+from associations.tests.base_test import BaseTestCase
+from authentication.models import User
+from repartitions.algorithm import get_project_index
+from repartitions.models import Proposition, Campaign, UserCampaign, Category, Wish
+
+
+class MunkresTestCase(BaseTestCase):
```
Piwakk: Munkres?

repartitions/tests/api.py
```diff
@@ -0,0 +1,386 @@
...
+
+    def test_user_endpoints(self):
+        # For an admin
+
+        self.login("17bocquet")
+
+        res = self.get("/repartitions/campaigns/1/users/")
+        self.assertEqual(res.status_code, 200)
+        self.assertEqual(
+            json.loads(res.content),
```
Piwakk: You can use `res.data` (same for everything below)

repartitions/algorithm.py
abocquet: Not necessarily, this is a pretty standard pattern

repartitions/models.py
abocquet: This is a one-to-many relationship so the ownership is defined on the `Choice` side

Piwakk: I mean, use `models.CharField(…, choices=STATUS)`

repartitions/permissions.py
```diff
@@ -0,0 +1,58 @@
+from rest_framework.permissions import BasePermission, SAFE_METHODS
+
+from authentication.models import User
+from repartitions.models import Wish, Campaign, Proposition, UserCampaign
+
+
+def user_in_campaign(user: User, campaign: Campaign) -> bool:
+    for category in campaign.categories.all():
+        if user in [uc.user for uc in category.users_campaign.all()]:
```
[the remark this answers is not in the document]
Piwakk: ping
I think it's more efficient at a database level

repartitions/algorithm.py
```diff
@@ -0,0 +1,248 @@
...
+        fixed_to = {}
+        for uc in user_campaigns:
+            if uc.fixed_to is not None:
+                fixed_to[uc.fixed_to.id] = fixed_to.get(uc.fixed_to.id, 0) + 1
+
+        # if it's not possible for the repartition to compensate the inequalities created by fixity
+        if len(fixed_to) > 0 and (max(fixed_to.values()) - 1) * len(propositions) > len(
+            user_campaigns
+        ):
+            raise Exception(
```
Piwakk: Still a broad `Exception` here, maybe a `RuntimeException`?

requirements.txt
```diff
@@ -1,25 +1,27 @@
...
 drf-extensions==0.5.0
 enum-compat==0.0.2
 isort==4.3.21
 lazy-object-proxy==1.4.3
 Markdown==3.1.1
 mccabe==0.6.1
-mypy==0.720
 mypy-extensions==0.4.3
+mypy==0.720
+nose==1.3.7
```
Piwakk: No need for nose any more right?

repartitions/permissions.py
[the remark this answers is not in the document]
abocquet: pong

PR #119, opened 2019-10-29
Abocquet/filter with tags

Piwakk:


associations/serializers/association.py
```diff
@@ -131,6 +132,22 @@ def get_my_role(self, obj):
         role = self.context["request"].user.get_role(obj)
         return RoleSerializer(role).data if role else {}
 
+    def get_pages(self, obj):
+        # Get the current user.
+        user = None
+        request = self.context.get("request")
+
+        if request and hasattr(request, "user"):
+            user = request.user
```
abocquet: maybe `user = getattr(self.context.get("request", {}), "user", none)`

associations/serializers/library.py
```diff
@@ -103,10 +105,21 @@ class LoanableSerializer(serializers.ModelSerializer):
...
+        read_only_fields = ("id", "tags")
+        fields = read_only_fields + (
+            "name",
+            "description",
+            "image",
+            "comment",
+            "library",
+        )
+
+    def get_tags(self, obj):
```
abocquet: Maybe create a `TagSerializerField` and make `tags = TagSerializerField` to avoid copy pasting the code in each serializer

associations/serializers/library.py
```diff
@@ -129,7 +142,43 @@ class Meta:
...
 
 class LibrarySerializer(serializers.ModelSerializer):
-    loanables = LoanableShortSerializer(many=True)
+    loanables = serializers.SerializerMethodField()
+
+    class Meta:
+        model = Library
+        fields = ("id", "enabled", "association", "loanables")
+
+    def get_loanables(self, obj):
```
abocquet: Same idea here

associations/views/library.py
```diff
@@ -37,13 +37,18 @@ def get_queryset(self):
...
+
+        return LibrarySerializer
+
 
 class LoanableViewSet(viewsets.ModelViewSet):
     queryset = Loanable.objects.all()
     serializer_class = LoanableSerializer
     permission_classes = (LoanablePermission,)
 
-    filter_backends = (DjangoFilterBackend,)
```
abocquet: Please keep this. It allows to make filtered queries from the rest API in the GET parameters

associations/views/library.py
```diff
@@ -66,7 +71,6 @@ class LoansViewSet(viewsets.ModelViewSet):
     queryset = Loan.objects.all()
     permission_classes = (LoansPermission,)
 
-    filter_backends = (DjangoFilterBackend, filters.OrderingFilter)
```
abocquet: Same

associations/views/marketplace.py
```diff
@@ -38,15 +38,18 @@ class MarketplaceViewSet(viewsets.ModelViewSet):
...
+
+        return MarketplaceSerializer
+
 
 class ProductViewSet(viewsets.ModelViewSet):
     queryset = Product.objects.all()
     serializer_class = ProductSerializer
     permission_classes = (ProductPermission,)
 
-    filter_backends = (filters.SearchFilter,)
```
abocquet: same

associations/views/marketplace.py
```diff
@@ -68,7 +71,6 @@ class TransactionViewSet(viewsets.ModelViewSet):
     serializer_class = TransactionSerializer
     permission_classes = (TransactionPermission,)
 
-    filter_backends = (DjangoFilterBackend,)
```
abocquet: same

associations/views/marketplace.py
```diff
@@ -131,7 +133,6 @@ class FundingViewSet(viewsets.ModelViewSet):
     serializer_class = FundingSerializer
     permission_classes = (FundingPermission,)
 
-    filter_backends = (DjangoFilterBackend,)
```
abocquet: same

authentication/views/profile.py
```diff
@@ -23,7 +23,6 @@ class ProfileViewSet(viewsets.ModelViewSet):
     serializer_class = UserSerializer
     permission_classes = (ProfilePermission,)
 
-    filter_backends = (SearchFilter, DjangoFilterBackend)
```
abocquet: same

authentication/views/questions.py
```diff
@@ -26,7 +26,6 @@ class ProfileAnswerViewSet(viewsets.ModelViewSet):
     serializer_class = ProfileAnswerSerializer
     permission_classes = (ProfileAnswerPermission,)
 
-    filter_backends = (SearchFilter, DjangoFilterBackend)
```
abocquet: same

abocquet: Hum I saw `DEFAULT_FILTER_BACKENDS ` has `django_filters.rest_framework.DjangoFilterBackend` so maybe it's ok removing it

associations/serializers/library.py
Piwakk: Well I tried, but the problem is that we need to have access to the `context` (aka the `user`) in this `Field`. And I don't see any obvious way to do it.
* The `Fields` do not seem to have a `context` variable automatically passed.
* We can always pass one to their constructor, but the place where the `Fields` are declared in a `Serializer` does not have access to the `context`.

abocquet: Maybe pass it as a constructor argument

authentication/views/questions.py
Piwakk: Indeed.

For the records, if `filter_backends` is declared in a `ViewSet`, the `HasHiddenTagFilter` will *not* work (unless explicitly written of course). So this line removal was actually to make it work.

As `DjangoFilterBackend` is used in almost every view, I think this is a good trade-of.
For `SearchFilter`, I think we will also end up using it everywhere anyway, so if there is no performance downside, we should put it as a default filter too.

associations/serializers/library.py
Piwakk: See comment above.

associations/serializers/association.py
Piwakk: OK, I think one line is a bit hard to follow, but you do have a point. How about:

```python
request = self.context.get("request", None)
user = getattr(request, "user", None)
```

PR #124, opened 2019-11-08
Remove `chat` from `backend`

Piwakk:


abocquet:
Tests don't pass 
```
  File "/Users/adrien.bocquet/Sites/Portail-des-eleves/backend/api/urls.py", line 6, in <module>
    import chat.urls
ModuleNotFoundError: No module named 'chat.urls'
```

PR #125, opened 2019-11-10
Set up authentication and 404 handling

abocquet:


frontend/src/App.tsx
```diff
@@ -1,18 +1,41 @@
...
 const App: React.FC = () => {
-    return (<Router>
-        <div className="App">
-            <Route path="/">
-            <Navbar/>
-            </Route>
-        </div>
-        </Router>
-    );
+    // Check the authentication then renders the page
```
Piwakk: render

frontend/src/App.tsx
```diff
@@ -1,18 +1,41 @@
...
+    let [checkedAuth, setCheckedAuth] = useState(false);
+    useEffect(() => {
+        authService.checkAuth().then(() => {
+            setCheckedAuth(true);
+        });
```
Piwakk: This effect is called at every render, consider specifying a condition for rendering

frontend/src/App.tsx
```diff
@@ -1,18 +1,41 @@
...
+                    {/* A PrivateRoute is a route that requires to be authenticated to be accessible.
+                     If the user is a public one, they'll be redirected to the login page. */}
+                    <PrivateRoute exact path="/" component={Homepage} />
+                    <Route path="/login" component={Login} />
+                    <Route component={PageNotFoundError} />
+                </Switch>
+            </Router>
+        );
+    } else {
+        return <p>Loading...</p>;
```
Piwakk: There is a <Spinner> component, might be interesting!

frontend/src/pages/homepage.tsx
```diff
@@ -0,0 +1,14 @@
+import React from "react";
+import Navbar from "../utils/navbar";
+import { PageTitle } from "../utils/common";
+
+export const Homepage = () => {
+    return (
+        <>
+            <Navbar />
```
Piwakk: I can smell that `Navbar` will be boilerplated everywhere, except in some specific pages.
Can we make an intermediate component (something like AuthenticatedUserContainer) for it?

frontend/src/App.tsx
Piwakk: `<Spinner>`*

frontend/src/pages/login.tsx
```diff
@@ -0,0 +1,48 @@
...
+                <div className="container">
+                    <div className="row">
+                        <div className="col col-login mx-auto">
+                            <div className="text-center mb-6 bg-white p-4 border rounded">
+                                <img src={Logo} alt="" />
+                            </div>
+                            <form className="card" action="" method="post">
+                                <div className="card-body p-6">
+                                    <div className="card-title">
+                                        Connection au portail des élèves
```
Piwakk: Connexion*
come on

frontend/src/pages/login.tsx
```diff
@@ -0,0 +1,48 @@
...
+export const Login = () => {
+    if (authService.isAuthenticated) {
+        return <Redirect to={"/"} />;
+    }
+
+    return (
+        <div className="page">
+            <div className="page-single">
+                <div className="container">
+                    <div className="row">
```
Piwakk: Boilerplate, right?
We must find a way to encapsulate this

frontend/src/utils/common.tsx
```diff
@@ -0,0 +1,12 @@
+import React from "react";
+
+/**
+ * This files should contain helpers for common components regularly used that
```
Piwakk: this file should contain helpers for regularly used components*

frontend/src/utils/privateRoute.tsx
```diff
@@ -0,0 +1,24 @@
+import React from "react";
+import { Route, Redirect } from "react-router-dom";
+import { authService } from "../App";
+
+/**
+ * Ensures the user is logged-in before displaying them the route
```
Piwakk: Ensure* even though this is controversial

frontend/src/utils/privateRoute.tsx
```diff
@@ -0,0 +1,24 @@
...
+ */
+export const PrivateRoute = ({ component: Component, ...rest }) => {
```
Piwakk: `rest` means the other props?

frontend/src/App.tsx
abocquet: image

abocquet: Try to add a `console.log("bip")` in the function and to navigate to different pages, it'll be called only once, because the app isn't re-rendered

frontend/src/App.tsx
abocquet: The user shouldn't even see it

frontend/src/App.tsx
Piwakk: OK, fine, au temps pour moi

frontend/src/pages/homepage.tsx
abocquet: Incoming in the next PR

frontend/src/App.tsx
Piwakk: Then remove the else in that case…
Sorry, but this is not very convincing x)
UI is also about details

frontend/src/pages/login.tsx
abocquet: Idk if the same css will be used in other places in the code. The login page is very specific

frontend/src/pages/homepage.tsx
Piwakk: what a cliffhanger

frontend/src/utils/privateRoute.tsx
abocquet: Yes

Piwakk: Is that a convention?
If not, would be better to call it `otherProps`
Yes, this is a detail

frontend/src/pages/login.tsx
Piwakk: OK fine, I confused with the ErrorPage

frontend/src/utils/privateRoute.tsx
abocquet: Apparenlty yes

PR #140, opened 2020-04-16
Chat client

Sheepsody:
Users can post messages on the chat

Added the following :
* frontend for the chat 
* API requests to retrieve the token
* Connect to the chat server and ask for the messages
* Updated the chat backend along with the associated tests


backend/manage.py
```diff
@@ -11,5 +11,5 @@
             "Couldn't import Django. Are you sure it's installed and "
             "available on your PYTHONPATH environment variable? Did you "
             "forget to activate a virtual environment?"
-        ) from exc
```
abocquet, line 14: Not sure this is inside the scope of the PR

chat_server/README.md
```diff
@@ -1,27 +1,44 @@
...
+
+```bash
+sudo -i -u postgres
+psql
+CREATE DATABASE portail_message OWNER root
+```
 
 ### Building the container 
 
+Another way to use the project is with containers. If you are using vscode, I would recommand checking the "Remote container" extension, but you can still run :
```
abocquet, line 23:
```suggestion
Another way to use the project is with containers. If you are using vscode, you should check the "Remote container" extension, but you can still run :
```

Prefer "you" to "I" when writing docs

cli/create_and_populate_database.bash
```diff
@@ -12,23 +12,23 @@ rm -Rf subscriptions/migrations/*
 rm -Rf tags/migrations/*
 
 
-python manage.py reset_db --noinput
+python3 manage.py reset_db --noinput
```
abocquet, line 15: The `python` command will be the right one when running the command in the virtual environment (`venv`). Please roll back

docker-compose.yml
```diff
@@ -0,0 +1,32 @@
...
+    # restart: always
+    # build:
+    #   context: ./chat_server/ 
+    #   dockerfile: chat_server.Dockerfile
+    # links:
+    #   - db:db
+    # env_file:
+    #    - ./chat_server/db.env
+    # ports:
+    #   - 3000:3000
```
abocquet, line 13: Should this section be activated ? If not maybe some clean up is possible. 
Anyway, I don't think the port is up to date (compared to `index.ts`)

docker/db.Dockerfile
```diff
@@ -0,0 +1,6 @@
+FROM postgres:latest
+
+COPY init-user-db.sh /usr/local/bin
+RUN ln -s /usr/local/bin/init-user-db.sh /docker-entrypoint-initdb.d/init-user-db.sh
+
+CMD ["postgres"]
```
abocquet, line 6: Maybe use the  basic `postgres` image here instead of defining a new docker image. The configuration you make in the `docker/init-user-db.sh` file is included in the postgres docker image (https://hub.docker.com/_/postgres)

docker/init-user-db.sh
```diff
@@ -0,0 +1,16 @@
+#!/bin/bash
+set -e
```
abocquet, line 2: See comment for `db.Dockerfile`

frontend/src/chat/Chat.tsx
```diff
@@ -0,0 +1,154 @@
...
+function createSocket() {
+    return api.jwt
+        .getToken()
+        .then((token: string) => {
+            return io.connect(chat_server_url, {
+                forceNew: true,
+                query: "token=" + token
+            });
+        })
+        .catch(() => {
```
abocquet, line 18: I don't think the `catch` is necessary  if you do nothing

docker-compose.yml
Sheepsody: I removed all changes related to the docker branch 
This branch could be use to deploy in the future, but it is not yet completely functional

frontend/src/chat/Chat.tsx
Sheepsody: Strange, it seems like our versions are different, any way this should be okay now

chat_server/README.md
Piwakk: And even `You can also use containers. With Visual Studio Code, you may use the "Remote container" extension; otherwise, you can still run:"

PR #155, opened 2020-05-09
Add the user promotion to the navbar

Piwakk:


Piwakk:
> 
> 
> Love it 🐴
> What about VS and ASTs?

The Portail is only for the _vrais Mineurs_.

Piwakk:
Merging for now, but we'll have to take the broad diversity of the _Mineurs_ into account => #156

PR #169, opened 2020-05-17
Create build-frontend.yaml

Piwakk:


.github/workflows/build-frontend.yaml
```diff
@@ -0,0 +1,27 @@
...
+  pull_request:
+      branches: [ master ]
+
+jobs:
+  build:
+
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Checkout
```
abocquet, line 13: Add cache to npm deps https://github.com/actions/cache/blob/master/examples.md#node---npm

PR #170, opened 2020-05-18
[avatarCard] Add `UserAvatarCard`

Piwakk:


frontend/src/components/utils/timeline/TimelineItem.tsx
```diff
@@ -28,8 +27,7 @@ export const TimelineItem = ({
         {content && content}
```
abocquet, line 27: Wouldn't `{content}` be enough?

Piwakk: Yes, definitely

PR #172, opened 2020-05-18
[fixWarnings] Fix warnings in `Chat` and `TagEdition`.

Piwakk:


frontend/src/chat/Chat.tsx
```diff
@@ -62,7 +62,7 @@ export const Chat = () => {
                 limit: 20,
             });
         })();
-    }, []);
+    }, [date]);
```
abocquet, line 65: The effect is bound to be executed only once

frontend/src/chat/Chat.tsx
```diff
@@ -87,7 +87,7 @@ export const Chat = () => {
                 scrollToLastMessage();
             });
         }
-    }, [socket, messages]);
+    }, [socket, messages, scrollToLastMessage]);
```
abocquet, line 90: Infinite loop

PR #177, opened 2020-05-21
Refactor user fields

Piwakk:
Please merge #173  before


backend/authentication/serializers/questions_short.py
```diff
@@ -1,12 +1,27 @@
...
     question = serializers.StringRelatedField()
+    question_id = serializers.SerializerMethodField()
+
+    def get_question_id(self, answer: ProfileAnswer):
+        return answer.question.id
 
     class Meta:
         model = ProfileAnswer
-        read_only_fields = ("id", "question", "text")
+        read_only_fields = ("id", "question_id", "question", "text")
```
abocquet: Don't we expect to have the `question` id in `question`?

frontend/src/components/user/edit_profile/EditUserProfile.tsx
```diff
@@ -0,0 +1,358 @@
...
+import * as Yup from "yup";
+import { SelectUsers } from "../../utils/forms/SelectUsers";
+
+// The items in the select.
+const currentAcademicYearItems = new Map([
+    ["1A", <span className="selectgroup-button">1A</span>],
+    ["2A", <span className="selectgroup-button">2A</span>],
+    ["GAP YEAR", <span className="selectgroup-button">Césure</span>],
+    ["3A", <span className="selectgroup-button">3A</span>],
+    ["GRADUATE", <span className="selectgroup-button">Alumni</span>],
```
abocquet: Maybe the span part is a bit redundant

backend/authentication/serializers/questions_short.py
Piwakk: Flat is better than nested

Piwakk: This is why I did it like this

Piwakk: question here is just a string

frontend/src/components/user/edit_profile/EditUserProfile.tsx
Piwakk: Yes, it's an issue with the component which use them… Let's keep it this way, I opened #179

PR #181, opened 2020-05-22
Manage association namespaces

abocquet:


frontend/src/components/associations/marketplace/ProductEdit.tsx
```diff
@@ -0,0 +1,167 @@
...
+    const newToast = useContext(ToastContext);
+    const history = useHistory();
+
+    const { data: product, status, error } = useBetterQuery<Product>(
+        ["product.get", productId],
+        api.products.get
+    );
+
+    const changeNumberLeftStatus = (numberLeft, setFieldValue) => {
+        const newNumberLeft = (numberLeft || 1) > -1 ? -1 : 1;
```
Piwakk, line 29: Please explain this

frontend/src/components/associations/marketplace/ProductEdit.tsx
```diff
@@ -0,0 +1,167 @@
...
+        <Loading className="mt-5" />
+    ) : error ? (
+        <Error detail={error} />
+    ) : product ? (
+        <Container>
+            <PageTitle>
+                <span onClick={history.goBack} className={"text-primary"}>
+                    <i className={"fe fe-arrow-left"} />
+                </span>
+                Editer le produit
```
Piwakk, line 43: modifier

frontend/src/components/associations/marketplace/ProductsAdministration.tsx
```diff
@@ -1,12 +1,44 @@
...
+import { Pagination } from "../../utils/Pagination";
+import { api } from "../../../services/apiService";
+import Row from "react-bootstrap/Row";
+import { Product } from "./common/Product";
+import { Link } from "react-router-dom";
 
 export const AssociationMarketplaceProductAdministration = ({
     association,
 }) => {
+    let marketplaceId = association.id;
```
Piwakk, line 12: cooonst

frontend/src/components/associations/marketplace/ProductEdit.tsx
```diff
@@ -0,0 +1,167 @@
...
+                }}
+                render={({ setFieldValue, values, errors }) => (
+                    <Form>
+                        <ReactBootstrapForm.Row className="mt-5">
+                            <Col md={{ span: 6 }}>
+                                <TextFormGroup
+                                    label="Nom du produit"
+                                    name="name"
+                                    type="text"
+                                    iconLeft="user"
```
Piwakk, line 78: I think `tag` would be a more appropriate icon.

frontend/src/components/associations/marketplace/ProductsAdministration.tsx
```diff
@@ -1,12 +1,44 @@
...
+                    render={(products, controlbar) => (
+                        <>
+                            <Row>
+                                {products.map((product) => (
+                                    <Product
+                                        product={product}
+                                        key={product.id}
+                                        additionalContent={
+                                            <Link
+                                                to={`/associations/${marketplaceId}/marketplace/products/${product.id}/edit`}
```
Piwakk, line 29: please rename the URL to `modifier` or something
Actually, `products` should be `produit`

Piwakk: And even `marketplace` should be translated

frontend/src/components/associations/settings/AssociationNamespaceSettings.tsx
```diff
@@ -0,0 +1,232 @@
...
+                        <InputGroup>
+                            <Form.Control
+                                placeholder={"Nouveau namespace"}
+                                name={"namespaceName"}
+                                value={values.namespaceName}
+                                onChange={handleChange}
+                                isInvalid={
+                                    (errors.namespaceName &&
+                                        touched.namespaceName) ||
+                                    undefined
```
Piwakk, line 121: why is there `undefined` here?
Doesn't it always evaluate to `false`? What's the point?

frontend/src/components/associations/settings/AssociationNamespaceSettings.tsx
```diff
@@ -0,0 +1,232 @@
...
+        onError: (errorAsUnknown) => {
+            const error = errorAsUnknown as AxiosError;
+            newToast({
+                message: `Erreur. Merci de réessayer ou de contacter les administrateurs si cela persiste. ${error.message}`,
+                level: ToastLevel.Error,
+            });
+        },
+    });
+
+    let type = tablerColors[hashCode(namespace.name) % tablerColors.length];
```
Piwakk, line 192: coooonst

frontend/src/components/associations/settings/AssociationNamespaceSettings.tsx
```diff
@@ -0,0 +1,232 @@
...
+            ) : (
+                <Modal show={true} onHide={closeModal}>
+                    <Modal.Header closeButton>
+                        <Modal.Title>Tags dans {namespace.name} </Modal.Title>
+                    </Modal.Header>
+                    <Modal.Body>
+                        {(data?.results.length || 0) > 0 ? (
+                            data?.results.map((tag) => (
+                                <span key={tag.id}>
+                                    <Tag type={type} tag={tag.value} />
```
Piwakk, line 213: Never paid attention to this before, but I feel like `type` should be called `color`.

frontend/src/components/utils/tags/TagEdition.tsx
```diff
@@ -100,7 +99,8 @@ export const TagEdition = ({ model, id }) => {
                     `Erreur durant le chargement : ${error.toString()}`
                 );
             });
-    }, [model, id]);
+        // eslint-disable-next-line
```
Piwakk, line 102: Please explain each time you are doing this
I don't understand why it happens so many times

frontend/src/routing/associations.ts
```diff
@@ -95,4 +98,17 @@ export const routes = (association) => [
         exact: true,
         props: { association: association },
     },
+    {
+        path: `/marketplace/products/:productId/edit`,
```
Piwakk, line 102: Make it Frenchier

frontend/src/components/associations/marketplace/ProductsAdministration.tsx
Piwakk: `magasin` is great

Piwakk:
I'm sorry, the following feedback will not always be about your PR. Your PR is already crossing borders between different topics, so I guess never mind.

image

Consider either use a small button (`btn-sm`) or increase the margins. Maybe both.

image

I would be happier with something which looks like https://preview.tabler.io/store.html. Either two or three columns of products (like on the left of Tabler's demonstration page), or a table of products (like on the right). Anyway, I think that the button is too big / visible compared to the other elements, and that the price should be more emphasized.

image

It's not 100 % clear that we can click on someone's name. I think adding an explanatory text (like “Cliquer sur quelqu’un pour modifier son compte / ses commandes” AND adding a `pointer` cursor (`cursor: pointer;` in CSS) would help.

image

I first think this lacks margins and better-adjusted alignments. Then, maybe redesign the organization of the form.

For instance, I think the balance should be more emphasized, and the `Credit account` input should only show up when clicking on a previous button (like in a modal, which shows an input and “shortcuts” buttons, like 5 €, 10 €, 50 €…)

Then, leverage Bootstrap's grid by splitting the form in two columns (the first row would be occupied by the balance and take the whole space).
First column, the items with a search input (just like what is currently done actually, but only in the left half).
Second column, the order.
Something like this (of course the margin and alignments still need to be adjusted, the “créditer” icône is ugly, etc.)
image
Still, I think it feels less cluttered and is thus more practical.
Also add pointers to the “remove item” button in the order.

image
The style should be the same as the “Magasin” style.

image
And this definitely needs to be in a card.
« Nombre restant » could be « Stock » it's more idiomatic I think.

Piwakk:
image

I think there is not enough content for each namespace to have its own card.

Maybe use a table here?

Also, adding a namespace does not automatically trigger the refetching of the namespaces.

frontend/src/components/associations/marketplace/ProductEdit.tsx
abocquet: This code is self-explanatory

abocquet: JK

frontend/src/components/associations/settings/AssociationNamespaceSettings.tsx
abocquet: `undefined || undefined == undefined`, I changed it to `(errors.namespaceName && touched.namespaceName) || false`

PR #196, opened 2020-05-24
Fix filtering of tags for taggable objects

abocquet:


backend/tags/views.py
```diff
@@ -20,6 +21,47 @@
...
+        return queryset
+
+    class Meta:
+        model = Namespace
+        fields = ("scoped_to_model", "scoped_to_pk")
+
+
+# for model in Tag.LINKED_TO_MODEL.keys():
+#     setattr(NamespaceFilter, model, NamespaceFilter._default_search_field)
+#     print(model)
```
Piwakk, line 62:
```suggestion
```

PR #200, opened 2020-05-24
Abocquet/clean marketplace

abocquet:


frontend/src/components/associations/marketplace/counter/CounterOrderMaker.tsx
```diff
@@ -103,38 +103,61 @@ export const CounterOrderMaker = ({
...
-                makeOrder={makeOrder}
-            />
+            <div className="d-flex justify-content-center mt-4">
+                <Col md={"6"}>
+                    <Card>
+                        <Card.Body className="text-center">
+                            <h5>{customer.id}</h5>
+                            <div className="display-2 font-weight-bold">
+                                {balanceStatus === "loading" ? (
+                                    <em>Chargement</em>
```
Piwakk, line 126: This is blown up because displayed in size `display-2` when loading
It would be better to have a Spinner, or something else more discreet. Nothing is good too.

frontend/src/components/associations/marketplace/counter/CounterOrderMaker.tsx
```diff
@@ -103,38 +103,61 @@ export const CounterOrderMaker = ({
...
+                    <ProductSearch
+                        basket={basket}
+                        addToBasket={addToBasket}
+                        marketplaceId={marketplaceId}
+                        compressed={Object.keys(basket).length > 0}
+                    />
+                </Col>
+
+                {/* Show order summary */}
+                {!basketIsEmpty && (
```
Piwakk, line 151: I think the UI should not change so much depending on `basketIsEmpty`.

Example: imagine you want to select the item 11 but select a wrong product (here item 13).

image

This is the screen when you are about to click on the red button to remove the item 13 from your order. And you already know the position of item 11, so you know where to click just after removing item 13.

But… the UI changes. And you might, in the end, select _another_ wrong item.

image

frontend/src/components/associations/marketplace/counter/OrderSummary.tsx
```diff
@@ -1,61 +1,72 @@
...
+        header: "Produit",
+        render: ({ product, quantity, status }) => (
+            <>
+                {product.name} <span className="text-muted">x{quantity}</span>
+            </>
+        ),
+    },
+    {
+        key: "value",
+        header: "Valeur",
```
Piwakk, line 47: image

I think « Montant » would be better.

There is also twice the quantity information (x2). Put a space between the « x » and the « 2 »

frontend/src/components/associations/marketplace/counter/CounterOrderMaker.tsx
```diff
@@ -3,7 +3,7 @@ import React, { useContext, useState } from "react";
 import { ToastContext, ToastLevel } from "../../../utils/Toast";
```
Piwakk, line 3: image


The arrow is a bit blown up here, and does not look like the other arrows used throughout the rest of the website. But I am not against this one specifically of course, we should just choose one version.

The arrow and the button « Recréditer le compte » are not aligned with the central card

frontend/src/components/utils/UserSelector.tsx
```diff
@@ -1,52 +1,61 @@
...
-            </>
-        );
-    }
-
-    return null;
+                    </>
+                )}
+            />
+            <p className="text-muted text-center">
+                Cliquez sur une carte pour séléctionner un élève
```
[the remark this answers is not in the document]
Piwakk: Also, please add the pointer cursor on the cards

frontend/src/components/associations/marketplace/counter/CounterOrderMaker.tsx
Piwakk: Also, please add the pointer cursor on the cards

frontend/src/components/associations/marketplace/ProductsAdministration.tsx
```diff
@@ -13,32 +14,70 @@ export const AssociationMarketplaceProductAdministration = ({
...
+                )}
+            />
         </>
     );
 };
+
+const columns = (marketplaceId) => [
+    {
+        key: "product",
+        header: "Product",
```
Piwakk, line 43:
```suggestion
        header: "Produit",
```

frontend/src/components/associations/marketplace/ProductsAdministration.tsx
```diff
@@ -13,32 +14,70 @@ export const AssociationMarketplaceProductAdministration = ({
...
+                      "restants"
+                  )}`
+                : "Stocks infinis",
+    },
+    {
+        key: "action",
+        header: "Actions",
+        render: (product) => (
+            <Link
+                to={`/associations/${marketplaceId}/magasin/produits/${product.id}/modifier`}
```
Piwakk, line 75: Would it be possible to have an icon button here?
I find the default Link style a bit sad. 😁

frontend/src/components/associations/marketplace/ProductEdit.tsx
```diff
@@ -14,6 +14,7 @@ import { TextFormGroup } from "../../utils/forms/TextFormGroup";
 import Button from "react-bootstrap/Button";
```
Piwakk, line 14: image

Please put the form inside a card.

Piwakk: When I click the « envoyer » button, and click on the arrow, I can't see my changes. I have to manually refresh the page to see them.

Piwakk: When writing a multiline description, it is not shown properly on the marketplace and on the table (maybe it's OK for the table).

image


image

Piwakk:
Extra comments on `Orders.tsx`

image

Should be « Statut » not « Status »
« Valeur » should be « Montant »

Why can the status be « Annulée »? I thought only the person who ordered could change it to « Annulée »
Also, instead of showing every choice, maybe show only the available choices. I am not sure, but I think in the backend, you cannot always go from one state to any other state… there is an order to be respected

abocquet:
There are many issues with data not reloading even if we call `queryCache.refetchQueries`. This is an issue across the whole application

abocquet:
About `Orders.tsx`, people in charge in the marketplace should be able to choose the state they deem more fit. Also in case of misclick it could be useful to come back.
The "Annulé" state is useful for an order made to the counter that is cancelled

frontend/src/components/associations/marketplace/QuantitySelect.tsx
```diff
@@ -2,30 +2,26 @@ import React, { useState } from "react";
...
-        return (
-            <Button variant="primary" onClick={increase}>
-                <i className="fe fe-plus" />
-                Commander
-            </Button>
-        );
-    }
-
-    return (
+    return quantity === 0 ? (
```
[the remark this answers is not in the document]
abocquet: It's big but it doesn't take space that should be attributed to something else so it's not an issue. Also it would be unnecessarily complicated to make an input of the length of 3 or 4 characters

frontend/src/components/associations/marketplace/Counter.tsx
```diff
@@ -9,10 +9,16 @@ import { CounterOrderMaker } from "./counter/CounterOrderMaker";
...
  */
 export const AssociationMarketplaceCounter = ({ association }) => {
-    const [customer, setCustomer] = useState<User | null>(null);
+    const [customer, setCustomer] = useState<User | null>({
+        firstName: "",
+        isStaff: false,
+        lastName: "",
+        promotion: "",
+        id: "17bocquet",
+    });
```
Piwakk, line 18:
```suggestion
    const [customer, setCustomer] = useState<User | null>(null);
```

PR #203, opened 2020-05-24
Marketplace search

abocquet:
Add search in marketplace

PR #205, opened 2020-05-28
Add useURLState hook for pagination

abocquet:
Will be soon used in tag search (when #200 and #203 are merged)


frontend/src/utils/useURLState.tsx
```diff
@@ -0,0 +1,44 @@
...
+        const params = new URL(window.location.href).searchParams;
+        value === defaultValue
+            ? params.delete(paramToWatch)
+            : params.set(paramToWatch, value);
+
+        history.push(location.pathname + "?" + params.toString());
+    };
+
+    useEffect(() => {
+        let params = new URLSearchParams(location.search);
```
Piwakk, line 31:
```suggestion
        const params = new URLSearchParams(location.search);
```

frontend/src/utils/useURLState.tsx
```diff
@@ -0,0 +1,44 @@
...
+
+        const paramJSONValue = params.get(paramToWatch);
```
Piwakk, line 33: Please explain what is going here. Why the need for JSON?

Piwakk:
Hi, and thanks for the work.

I noticed a behaviour which is not desirable.

For the purpose of testing, I asked the backend to return pages with size 2. Here is what I get in medias:

image

Now, if I click on the checkboxes in the sidebar, the frontend will call `/api/v1/associations/media/?tags__are=1,2,3,4&association=bde&page=4&page_size=2` (for instance), which returns a 404. So, nothing changes, and this is not a good UX.

Piwakk:
Also, when leaving and going back to the page, the checkboxes previously checked uncheck themselves. I think this has to do with react-query refetching the page on focus, can you confirm?

abocquet:
This is not a bug. It's a feature.

🚀

Piwakk:
> 
> 
> This is not a bug. It's a feature.
> 
> 🚀

You are so pragmatic! ❤️

frontend/src/components/utils/Pagination.tsx
```diff
@@ -50,14 +51,33 @@ export const Pagination = ({
...
-    const history = useHistory();
-    const location = useLocation();
-    const [page, setPage] = useState(1);
+    const [page, setPage] = useURLState("page", 1);
     const [maxPage, setMaxPage] = useState(1);
-    const [parsedLocation, setParsedLocation] = useState(false);
+
+    // We have to use this "temporized" field to make sure we always update the
+    // page before the parameters, in  order to avoid unnecessary requests
+    // which might end up in 404
```
Piwakk, line 59:
```suggestion
    // Before propagating a change of `apiKey`, we reset the page to 1.
    // Indeed, if the `apiKey` induces a change of the maximum available page
    // (e.g. if `apiKey` contains a filter), we don't want the “old” page prop
    // to exceed this maximum.
    // However, changing the page may trigger a rerendering of the component
    // using `Pagination`, which, in turn, may change the `apiKey` prop, losing the
    // initial key. Hence we save this “initial” key in `temporizedApiKey`.
```

PR #214, opened 2020-05-31
Table for roles management frontend

Piwakk:
Only the table of the members administration page + sidebar for future pages (dead links for now).

(Go to http://localhost:3000/associations/bde/membres/administration.)


frontend/src/components/associations/roles/administration/AssociationRolesAdministration.tsx
```diff
@@ -0,0 +1,148 @@
...
+
+export const AssociationRolesAdministration = ({
+    association,
+}: {
+    association: Association;
+}) => {
+    // The Role currently edited in the modal.
+    const [, setEditRole] = useState<Role | null>(null);
+
+    const columnData = [
```
abocquet, line 59: It may be more readable to always  but column data outside of the components

PR #220, opened 2020-06-01
Add role icons

Piwakk:
Add some components which will be used in an upcoming PR.

### DIFF
--- a/frontend/src/components/associations/roles/administration/RolePermissionIconTooltip.tsx
+++ b/frontend/src/components/associations/roles/administration/RolePermissionIconTooltip.tsx
@@ -1,0 +1,22 @@
+import React from "react";
+import { RolePermission } from "../../../../models/associations/role";
+import { OverlayTriggerProps } from "react-bootstrap";
+import { RolePermissionTooltip } from "./RolePermissionTooltip";
+import { RolePermissionIcon } from "./RolePermissionIcon";
+
+export const RolePermissionIconTooltip = ({
+    permission,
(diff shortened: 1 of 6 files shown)


frontend/src/components/associations/roles/administration/RolePermissionIconTooltip.tsx
```diff
@@ -0,0 +1,22 @@
...
+    permission: RolePermission;
+    iconProps?: React.DetailedHTMLProps<
+        React.HTMLAttributes<HTMLElement>,
+        HTMLElement
+    >;
+    tooltipProps?: Partial<OverlayTriggerProps>;
+}) => (
+    <RolePermissionTooltip permission={permission} {...tooltipProps}>
+        <RolePermissionIcon permission={permission} {...iconProps} />
+    </RolePermissionTooltip>
```
abocquet, line 21: Why did we need to split `RolePermissionIcon` into `RolePermissionTooltip` and `RolePermissionIcon` knowing they are always used in pair? Doesn't it make the code over-layered?

Piwakk: I will use them separately in a further PR

Piwakk: Where `RolePermissionIcon` will be wrapped into something (a select) and I want the tooltip to be applied to the whole select, not the icon

PR #221, opened 2020-06-01
Refactor `TextFormGroup`

Piwakk:
Better typing and documentation + base component for further refactoring.


Piwakk:
Yes, it's a base component which will be useful for further refactoring

PR #226, opened 2020-06-03
Add CreateRole

Piwakk:
image

Add a Modal built on the previous one.


Piwakk:
> 
> 
> It should not be possible to create two roles for the same user
> image
> 
> For another PR : it would be nice to be able to filter by still active (or not) permissions

I think this should be allowed. You could totally have one role during the first year and another during the second year; this happened to me

abocquet:
Such a _cumular_

PR #227, opened 2020-06-04
Use url params in text search and tag search

abocquet:
Merge **after** #230


frontend/src/utils/useURLState.tsx
```diff
@@ -37,9 +41,13 @@ export function useURLState<T>(
         // Convert the object to a string using JSON because we don't know
         // what type of data we're working with in advance. Also helps to
         // cast non string value. For instance, "2" will become the number 2
-        const paramJSONValue = params.get(paramToWatch);
-        if (paramJSONValue !== null) {
-            setParamValue(JSON.parse(paramJSONValue) as T);
+        const urlParamValue = params.get(paramToWatch);
+        if (urlParamValue !== null) {
+            // TS isn't able to see that stateIsString true implies urlParamValue is a string
```
Piwakk, line 46: I don't understand why it should be too.

Piwakk: I would be interested in an explanation
If it works, I think a typeguard would be safer and more explicit: `urlParamValue as string`

frontend/src/components/utils/tags/TagSearch.tsx
```diff
@@ -7,92 +7,159 @@ import { SidebarSection } from "../sidebar/SidebarSection";
...
+ *     tagsQueryParams={{
+ *         page_size: 1000,
+ *         namespace__scoped_to_model: "association",
+ *         namespace__scoped_to_pk: associationId,
+ *         related_to: "product",
+ *     }}
+ * />
+ * will display a tag search form for all products related to the association
+ * with ID `associationId`. Note the page_size is fixed to 1000 because it
+ * allows fluid UI but there never be more than a thousand tags for an association.
```
Piwakk, line 27:
```suggestion
 * allows fluid UI but there will never be more than a thousand tags for an association.
```

frontend/src/components/utils/tags/TagSearch.tsx
```diff
@@ -7,92 +7,159 @@ import { SidebarSection } from "../sidebar/SidebarSection";
...
+ * This is also the upper limit for the backend, so don't forget to change
+ * both the frontend and the backend if you want to increase it.
+ *
+ * @param tagsQueryParams a literal object of filtering parameters we want to use
+ * to query the tags that will be proposed to the user. Useful to get only tags
+ * related to a particular scope
+ * @param setParams a `useState` variable setter in which the additional object
+ * filtering parameters will be put. You'll simply have to unwrap them in
+ * `getBetterPaginatedQuery` to use them
```
Piwakk, line 36: You mean `toUrlParams`?
Maybe put an example of how to use the `tagParams`

PR #229, opened 2020-06-04
Change all AND tag search for AND and OR (cf issue #208)

### LINKED ISSUE #208: Sidebar checkboxes of a same section should be ORs, not ANDs
Sidebar checkboxes of a same section are currently doing an AND. However, the user expects it to do an OR on the conditions.

Currently:

image

But for instance in this website:

image

abocquet:
Fix #208

### DIFF
--- a/backend/tags/filters/taggable.py
+++ b/backend/tags/filters/taggable.py
@@ -29,9 +29,7 @@ class TaggableFilter(filters.FilterSet):
                 conditions = dict()
                 # regroup the tags by namespace, in a "tag id in" condition
                 for tag in tags:
-                    conditions[tag["namespace"]] = conditions.get(
-                        tag["namespace"], []
-                    ) + [tag["id"]]
+                    conditions.setdefault(tag["namespace"], []).append(tag["id"])
 
                 # regroup the conditions in a AND condition with multiple filtering
                 for key in conditions.keys():
--- a/backend/tags/filters/taggable.py
+++ b/backend/tags/filters/taggable.py
@@ -1,6 +1,10 @@
+from functools import reduce
+
 import django_filters
 from django.db.models import Q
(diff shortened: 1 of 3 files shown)


backend/tags/filters/taggable.py
```diff
@@ -20,9 +24,18 @@ def tags_are(self, queryset, value, *args, **kwargs):
...
-                    new_condition = Q(**{"tags__id": tag_id})
-                    queryset = queryset & queryset.filter(new_condition)
+                tags = Tag.objects.filter(id__in=tag_ids).values("id", "namespace")
+
+                conditions = dict()
+                # regroup the tags by namespace, in a "tag id in" condition
+                for tag in tags:
+                    conditions[tag["namespace"]] = conditions.get(
+                        tag["namespace"], []
+                    ) + [tag["id"]]
```
Piwakk, line 34:
```suggestion
                    conditions.setdefault(tag["namespace"], []).append(tag["id"])
```

backend/tags/filters/taggable.py
```diff
@@ -1,7 +1,11 @@
+from functools import reduce
+
```
Piwakk, line 2:
```suggestion
```

backend/tags/filters/taggable.py
```diff
@@ -1,7 +1,11 @@
...
 import django_filters
 from django.db.models import Q
 from django_filters import rest_framework as filters
 
+from tags.models import Tag
+
 """
 Taggable filter allows to filter taggable objects with an intersection of tags. If you're working with Media
```
Piwakk, line 10: Update the comment

Piwakk:
When I select tags then switch tabs, the checkboxes uncheck themselves (but the filter is still applied).

Before switching tab:
image

After:
image

PR #234, opened 2020-06-07
Refactor SelectFormGroup

Piwakk:


backend/authentication/tests/test_birthdays.py
```diff
@@ -40,7 +40,7 @@ def test_if_not_logged_in_then_401(self):
 
     def test_birthdays(self):
         self.load_birthdays()
-        self.login("17simple")
+        self.login("user0")
```
abocquet, line 43: Use a more realistic username, like "18birthdays"

PR #237, opened 2020-06-07
Split requirements.txt in two versions: dev and prod.

Piwakk:


Piwakk:
The Django tests do not pass, but that's because I haven't merged with #236

PR #252, opened 2020-06-20
Piwakk/library frontend

Piwakk:
Add loan / cancel buttons to `LoanableCard`.


backend/associations/serializers/library.py
```diff
@@ -121,10 +122,29 @@ class Meta:
...
+
+    def get_number_of_pending(self, loanable):
+        """Return the number of pending requests for this loanable."""
+        return len(loanable.loans.filter(status="PENDING"))
+
     def to_representation(self, instance: Loanable):
         res = super().to_representation(instance)
 
-        res["status"] = "AVAILABLE" if instance.is_available() else "BORROWED"
+        res["status"] = instance.status
```
abocquet, line 147: Why can't this go in the `read_only_fields` ?

frontend/src/components/associations/library/home/LoanableCard.tsx
```diff
@@ -1,48 +1,158 @@
...
+  );
+
+  const UnavailableButton = () => (
+    <Button className="btn-sm" variant="outline-danger" disabled={true}>
+      Indisponible
+    </Button>
+  );
+
+  if (loanable.status === "BORROWED") {
+    return <UnavailableButton />;
```
abocquet, line 54: `return (status === "BORROWED") : <case 1> ? <other cases>`

frontend/src/components/associations/library/home/LoanableCard.tsx
```diff
@@ -1,48 +1,158 @@
...
+      placement="bottom"
+      overlay={<Tooltip id={`${loanable.id}-tooltip`}>{tooltipText}</Tooltip>}
+    >
+      <Button className="btn-sm" variant="primary" onClick={loan}>
+        Emprunter
+      </Button>
+    </OverlayTrigger>
+  );
+
+  const UnavailableButton = () => (
```
abocquet, line 47: Maybe not useful to define the component to call it only once after that. It would be simpler to directly inject the code where you use the tag

Piwakk: It's more explicit this way and does not break the symmetry with the other cases

frontend/src/components/associations/library/home/LoanableCard.tsx
Piwakk: Yes, I corrected it in #256

PR #253, opened 2020-06-20
Piwakk/library frontend management

Piwakk:
Add edit form for loanables.
Should review after #252 
Filters will be added next.

image

image


Piwakk:
#253

PR #256, opened 2020-06-20
Piwakk/library frontend filters

Piwakk:
Requires #253


backend/associations/views/library.py
```diff
@@ -44,13 +49,71 @@ def get_serializer_class(self):
...
+
+    def filter_status(self, queryset, _, filter):
+        """Filter by AVAILABLE, BORROWED or REQUESTED.
+                                   | The loanable is available | The loanable is borrowed |
+        There are PENDING loans    | REQUESTED                 | Impossible               |
+        There are no PENDING loans | AVAILABLE                 | BORROWED                 |
+        """
+
+        # No filter or every filter.
+        if len(filter) == 0 or len(filter) == 3:
```
abocquet, line 74: What if `filter` has 2 invalid values and 1 valid value ?  (ex: `[AVAILABLE, PICHE, CLAC]`

backend/associations/views/library.py
```diff
@@ -44,13 +49,71 @@ def get_serializer_class(self):
...
+        ids = set()
+
+        if "AVAILABLE" in filter:
+            ids.update(
+                [
+                    x.id
+                    for x in queryset
+                    if x.is_available() and x.number_of_pending_loans == 0
+                ]
+            )
```
abocquet, line 86: Replace with additional filters

backend/associations/views/library.py
```diff
@@ -44,13 +49,71 @@ def get_serializer_class(self):
...
+
+        if "BORROWED" in filter:
+            ids.update([x.id for x in queryset if not x.is_available()])
```
abocquet, line 89: Replace with additional filters

backend/associations/views/library.py
```diff
@@ -44,13 +49,71 @@ def get_serializer_class(self):
...
+
+        if "REQUESTED" in filter:
+            ids.update(
+                [
+                    x.id
+                    for x in queryset
+                    if x.is_available() and x.number_of_pending_loans > 0
+                ]
```
abocquet, line 97: Replace with additional filters

frontend/src/components/associations/library/home/AssociationLibraryHome.tsx
```diff
@@ -8,38 +8,123 @@ import Row from "react-bootstrap/Row";
...
+              state={
+                statusParams.status.includes("AVAILABLE") &&
+                statusParams.status.includes("REQUESTED")
+              }
+              onChange={(ticked) =>
+                setStatusParams((oldStatus) => ({
+                  status: ticked
+                    ? [...oldStatus.status, "AVAILABLE", "REQUESTED"]
+                    : oldStatus.status.filter(
+                        (x) => x !== "AVAILABLE" && x !== "REQUESTED"
```
abocquet, line 59: If I remember well, on there is somewhere a `priority` mixin because we can potentially have multiple requests for the same Loanable at the same time. But here a Loanable is considered unavailable as soon it's requested, so the backend and the frontend are not on the same page here

frontend/src/models/associations/library.ts
```diff
@@ -19,13 +19,13 @@ export interface Loan {
...
   userLoan: null | (Omit<Loan, "loanable"> & { loanable: number });
-  numberOfPending: number;
+  numberOfPendingLoans: number;
   name: string;
   description?: string;
   image?: string;
   comment?: string;
   library: Library;
-  status: "AVAILABLE" | "BORROWED" | "REQUESTED";
+  status: "AVAILABLE" | "BORROWED";
```
abocquet, line 28: Won't `REQUESTED` be able with the admin UI ?

backend/associations/serializers/library.py
```diff
@@ -137,13 +137,16 @@ def get_user_loan(self, loanable):
...
-        return len(loanable.loans.filter(status="PENDING"))
+    def get_number_of_pending_loans(self, loanable):
+        """Return the number of pending loans for this loanable."""
+        return loanable.number_of_pending_loans
+
+    def to_representation(self, loanable: Loanable):
+        res = super().to_representation(loanable)
+
+        res["status"] = "AVAILABLE" if loanable.is_available() else "BORROWED"
+        res["expected_return_date"] = loanable.get_expected_return_date()
```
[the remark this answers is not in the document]
Piwakk: Not sure I understood your remark, but you want to know, at the loanable level, when it will be returned, without having to dig into the loans

backend/associations/views/library.py
Piwakk: It's verified before (see `MultipleChoiceFilter`)

backend/associations/views/library.py
Piwakk: This filter relies on a property is_available, so it cannot be filtered in a Queryset.filter

frontend/src/components/associations/library/home/AssociationLibraryHome.tsx
Piwakk: It's the opposite here: keeping the “Disponible” loanables will send a request for filtering on the AVAILABLE and REQUESTED loanables

frontend/src/models/associations/library.ts
Piwakk: See `AssociationLibraryManagement`

Piwakk: Maybe one confusing thing here should be explained:
- The data, whether in the backend or the frontend side, only has the `AVAILABLE` and `BORROWED` status (you can see this in `library.ts` and in `models/library.py`).
- However, the _filter_ does not use the same convention (which is confusing I admit): you have three values, `AVAILABLE`, `BORROWED` and `REQUESTED`, to help separating the “available but requested” and “available but not requested” states, useful for the admins.

backend/associations/views/library.py
abocquet: It can,  it's just not built in for now

backend/associations/serializers/library.py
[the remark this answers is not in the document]
abocquet: Yes, but here we're looking at `Loan`. Let's say you  want to know it a loanable was given back late. You'll look at `expected_return_date` and `actual_return_date`, but since `expected_return_date` is not the Loan one but the Loanable one, you'll never be able to get that information

frontend/src/models/associations/library.ts
abocquet: Write a comment then please

backend/associations/serializers/library.py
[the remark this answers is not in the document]
Piwakk: > Yes, but here we're looking at Loan.

This is a `Loanable` serializer. I don't understand your point

> Let's say you want to know it a loanable was given back late. You'll look at expected_return_date and actual_return_date, but since expected_return_date is not the Loan one but the Loanable one, you'll never be able to get that information

Yes in that case you look at the `Loan`, not at the `Loanable`.

frontend/src/models/associations/library.ts
Piwakk: Well it's explained in the filters… and in the UI when necessary

Piwakk: But this could be better I'll change it

Issue #255, opened 2020-06-20
Add badges in the sidebars indicating the number of pending items


Piwakk:
For the administration pages. Something like this:
![image](https://user-images.githubusercontent.com/11032610/85211440-85620700-b349-11ea-961b-da5aae3ba215.png)

PR #257, opened 2020-06-20
Piwakk/add polls badges

Piwakk:
Add a generic component for the sidebar badges and use it in Polls.
Addresses _part_ of #255 

image

### DIFF
--- a/backend/polls/views.py
+++ b/backend/polls/views.py
@@ -2,7 +2,7 @@ from datetime import date
 
 from django.db.models import Q
 from django_filters.rest_framework import FilterSet, CharFilter, MultipleChoiceFilter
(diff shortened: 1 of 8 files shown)


backend/polls/views.py
```diff
@@ -92,6 +92,28 @@ def results(self, *args, **kwargs):
...
+         - the number of polls to which the user can vote."""
+        return response.Response(
+            data={
+                "number_of_pending_polls": len(Poll.objects.filter(state="REVIEWING"))
+                if self.request.user.is_staff
+                else None,
+                "number_of_available_polls": len(
+                    [
+                        poll
+                        for poll in Poll.objects.filter()
```
abocquet, line 108: We could get better perfs by actually using the `filter` method and the `count()`

PR #262, opened 2020-06-23
Media preview for PDFs

abocquet:
Same as #261 but for pdfs 

image


backend/associations/models/media.py
```diff
@@ -30,6 +39,51 @@ class Meta:
...
+            image = Image.open(instance.file.path)
+            image.thumbnail((300, 300))
+            image.save(output, format=extension)
+            instance.preview.save(
+                os.path.splitext(filename.name)[0] + "_preview" + filename.suffix,
+                ContentFile(output.getvalue()),
+            )
+
+            instance.save()
+
```
Piwakk, line 77:
```suggestion
```

PR #274, opened 2020-07-04
[Library] Add filters for loans management

Piwakk:


backend/associations/views/library.py
```diff
@@ -123,11 +123,36 @@ def get_queryset(self):
...
+        method="filter_status",
+    )
+
+    class Meta:
+        model = Loan
+        fields = ("user", "loanable__id", "loanable__library", "status")
+
+    def filter_status(self, queryset, _, filter):
+        # No filter or every filter.
+        if len(filter) == 0 or len(filter) == 6:
```
abocquet, line 145:
```suggestion
        if len(filter) == 0 or len(filter) == len(Loan.STATUS):
```

PR #278, opened 2020-07-06
Rewrite the elections backend

Piwakk:


backend/associations/serializers/election.py
```diff
@@ -1,83 +1,97 @@
...
+        model = Choice
+        fields = ("name", "number_of_offline_votes")
 
+    def to_representation(self, instance: Choice):
+        response = super(ElectionChoiceSerializer, self).to_representation(instance)
 
-class ElectionSerializer(serializers.ModelSerializer):
-    """This serializer gives access to the registered_voters and choices fields.
+        if instance.election.show_results:
+            response["number_of_online_votes"] = instance.number_of_online_votes
```
abocquet, line 69: Why always show `number_of_offline_votes` but sometimes `number_of_online_votes` ?

backend/associations/serializers/election.py
```diff
@@ -114,34 +128,9 @@ def validate_against_instance(cls, instance, validated_data):
...
-            # Add the new choices not yet in the Choice table.
-            Choice.objects.bulk_create(
-                [
-                    Choice(election=election, name=choice_name)
-                    for choice_name in new_choices_names - old_choices_names
-                ]
-            )
+        for field in ("association", "voters", "choices"):
+            if field in validated_data:
+                validated_data.pop(field)
```
abocquet, line 133: Maybe they should be in `read_only_fields` then ? (at least for `voters` and `choices` since they are removed during update and create)

backend/associations/serializers/election.py
```diff
@@ -1,83 +1,97 @@
...
+    """This serializer is not bound to any model and is used for the /vote/ action."""
+
+    choices = serializers.PrimaryKeyRelatedField(
+        queryset=Choice.objects.all(), many=True, read_only=False
+    )
+
+    def create(self, validated_data):
+        raise NotImplementedError(
+            "This serializer cannot be used for database operations."
+        )
```
abocquet, line 19: Why don't you use the same pattern as in `ChoiceReadOnlySerializer` ?

backend/associations/serializers/election_read_only.py
```diff
@@ -0,0 +1,53 @@
...
+            "user_voter",
+        )
+        fields = read_only_fields
+
+    def save(self, **kwargs):
+        raise NotImplementedError("This serializer is read-only.")
+
+    def get_user_voter(self, election):
+        query = election.voters.filter(user=self.context["request"].user)
+        return VoterSerializer(query[0]).data if query.exists() else None
```
abocquet, line 53:
```suggestion
        res = election.voters.filter(user=self.context["request"].user).get()
        return VoterSerializer(res).data if res else None

```

backend/associations/tests/election/test_choice.py
```diff
@@ -0,0 +1,306 @@
...
+
+    def test_new_number_of_online_votes_is_zero(self):
+        self.login("17election_pdm")
+        self.assertFalse(Election.objects.get(pk=20).started)
+
+        res = self.create(
+            {
+                "election": 20,
+                "name": "Un choix",
+                "number_of_online_votes": 4,
```
abocquet, line 136: Shouldn't `number_of_online_votes` be read-only (but updated by the views) ?

backend/associations/views/election.py
```diff
@@ -1,92 +1,220 @@
...
+                "The status cannot be updated as the election has not started."
+            )
+
+        # The status can only be updated from PENDING.
+        if self.get_object().status != "PENDING":
+            raise ValidationError("The status cannot be updated anymore.")
+
+        # The status can only be updated to OFFLINE_VOTE (as ONLINE_VOTE is automatically handled).
+        if serializer.validated_data["status"] != "OFFLINE_VOTE":
+            raise ValidationError("The status can only be updated to OFFLINE_VOTE.")
```
abocquet, line 65:
```suggestion
            raise ValidationError("A vote created by the staff can only have the status OFFLINE_VOTE.")
```

backend/associations/views/election.py
```diff
@@ -1,92 +1,220 @@
...
-"""
+        serializer.save()
+
+    def perform_update(self, serializer):
+        choice = self.get_object()
+        override = {"election": choice.election}
+
+        # If the election has started, do not change the name.
+        if self.get_object().election.started:
+            override["name"] = choice.name
```
abocquet, line 117: else send 403 ou 400

backend/associations/views/election.py
```diff
@@ -1,92 +1,220 @@
...
+            else {}
+        )
 
-    def create(self, request, *args, **kwargs):
-        election = self.get_election_or_404(**kwargs)
+        serializer.save(
+            association=election.association,
+            choices=election.choices,
+            voters=election.voters,
+            **override
```
abocquet, line 164: Why do we need to override ?

abocquet: These fields will be let alone be  default, won't they ?

backend/associations/views/election.py
```diff
@@ -1,92 +1,220 @@
...
-        election.voters.add(request.user)
+        # Update the voter state.
+        voter = election.voters.get(user=self.request.user, status="PENDING")
+        voter.status = "ONLINE_VOTE"
+        voter.save()
+
+        # Update the choices.
+        for choice in serializer.validated_data["choices"]:
+            choice.number_of_online_votes += 1
+            choice.save()
```
abocquet, line 215: Should be done in a single SQL transaction 
https://docs.djangoproject.com/fr/2.2/topics/db/transactions/

backend/associations/serializers/election.py
Piwakk: Because nobody should see the results of the election until `show_results` is True, including the election administrator
However, the election administrator can always edit / see `number_of_offline_votes`

backend/associations/serializers/election.py
Piwakk: They are used in create (but not in update).

backend/associations/tests/election/test_choice.py
Piwakk: Yes, but that's the point of the test: `test_new_number_of_online_votes_is_zero`

backend/associations/views/election.py
Piwakk: You mean if the election has not started?

backend/associations/views/election.py
Piwakk: The serializer currently used does pop these fields during the create, but it's no harm enforcing this rule here too

backend/associations/serializers/election.py
Piwakk: The same pattern? What do you mean?

backend/associations/views/election.py
Piwakk: But if the election has not started we can still edit

backend/associations/views/election.py
Piwakk: Yes… I really don't understand what you want to change here sorry

`override` is either `{}` or `{ name: ..., ... }` depending on `election.started`

Piwakk:
For the record: https://github.com/actions/virtual-environments/issues/1202

PR #288, opened 2020-08-02
Choices form + election deletion

abocquet:


backend/associations/serializers/election.py
```diff
@@ -55,16 +55,13 @@ class Meta:
...
-            response["number_of_votes"] = instance.number_of_votes
-
-        return response
+        fields = (
+            "id",
+            "name",
+            "number_of_offline_votes",
+            "number_of_online_votes",
+            "number_of_votes",
+        )
```
abocquet, line 64: Before you ask, this is necessary to allow the organiser see the results before they publish them

PR #293, opened 2020-09-05
Piwakk/association admin settings

Piwakk:


abocquet (approved):
Maybe we should rename "membres" in "utilisateurs" or so, because "membres" could refer to the association's members, which is a bit disturbing 
image

abocquet (approved):

PR #297, opened 2020-09-19
Victor/stats sondage

victor-amblard:
Basic features to have a couple stats about poll results by user...

### DIFF
--- a/backend/polls/views.py
+++ b/backend/polls/views.py
@@ -26,6 +26,37 @@ from polls.permissions import (
 )
 
 
+def generate_stats():
+    # Computing the rankings...
+    all_users = User.objects.all()
+    all_polls = Poll.objects.all()
+    poll_leaderboard = {}
+
+    for user in all_users:
+        poll_leaderboard[user.id] = [0, 0, 0]
+
+    for poll in all_polls:
+        if not poll.has_been_published:
+            continue
+
+        poll_date = poll.publication_date
+        coeff_poll = exp(-(date.today() - poll_date).days / 14)
+        voters = poll.votes.values_list("user__id", "choice")
+
+        for voter in voters:
+            (user_id, user_choice) = voter
+            flag_victory = (user_choice == poll.results) + 1  # 2 if won, 1 otherwise
+
+            poll_leaderboard[user_id][flag_victory] = (
+                coeff_poll + poll_leaderboard.get(user_id, 0)[flag_victory]
+            )
+            poll_leaderboard[user_id][0] = (
+                1 + poll_leaderboard.get(user_id, 0)[0]
+            )  # participations
+
+    return poll_leaderboard
+
+
 class PollFilter(FilterSet):
     """
     This class is needed because Django does not allow to filter on properties.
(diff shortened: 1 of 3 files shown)


backend/polls/views.py
```diff
@@ -118,6 +128,104 @@ def stats(self, *args, **kwargs):
...
+            participations_by_user = cache.get("participations")
+            victories_by_user = cache.get("victories_ranking")
+            defeats_by_user = cache.get("defeats_ranking")
+
+        else:
+            # Computing the rankings...
+
+            for poll in self.queryset:
+                if poll.has_been_published:
+                    pollDate = poll.publication_date
```
abocquet, line 155:
```suggestion
                    poll_date = poll.publication_date
```
We try to use `snake_case` for python and `camelCase` for TS

backend/polls/views.py
```diff
@@ -118,6 +128,104 @@ def stats(self, *args, **kwargs):
...
+
+                    for voter in voters:
+                        user_id = voter[0]
+                        user_choice = voter[1]
+
+                        if user_choice == poll.results:
+                            if user_id in victories_by_user:
+                                victories_by_user[user_id] += coeff_poll
+                            else:
+                                victories_by_user[user_id] = coeff_poll
```
abocquet, line 167:
```suggestion
                           victories_by_user[user_id] = coeff_poll + victories_by_user.get(user_id, 0)
```

backend/polls/views.py
```diff
@@ -118,6 +128,104 @@ def stats(self, *args, **kwargs):
...
+                        else:
+                            if user_id in defeats_by_user:
+                                defeats_by_user[user_id] += coeff_poll
+                            else:
+                                defeats_by_user[user_id] = coeff_poll
+
+                        if user_id in participations_by_user:
+                            participations_by_user[user_id] += 1
+                        else:
+                            participations_by_user[user_id] = 1
```
abocquet, line 177: Same thing

backend/polls/views.py
```diff
@@ -118,6 +128,104 @@ def stats(self, *args, **kwargs):
...
+
+        sorted_users_participations = sorted(
+            participations_by_user.items(), key=itemgetter(1)
+        )[::-1]
+        sorted_users_defeats_floating = sorted(
+            defeats_by_user.items(), key=itemgetter(1)
+        )[::-1]
+        sorted_users_victories_floating = sorted(
+            victories_by_user.items(), key=itemgetter(1)
+        )[::-1]
```
abocquet, line 216: Maybe the ranking should be limited to the top 10, 20 or 30 people

abocquet (changes requested):
The code tend to mix up cache setting and score calculation. The cache is a layer above the calculation function. To clear things up maybe you could create a function that calculates everything, without caching, and then cache it if needed 

```
def make_statistics():
    do_stuff()

class PollViewset:
    ....
    def statistics(self):
         if "leaderboard" not in  cache or "user_scores" not in cache:
            leaderboard, user_scores = make_statistics()
            cache.set("leaderboard", leaderboard)
            cache.set("user_scores", user_scores)
        return Response(data={
             "leaderboard": cache.get("leaderboard"), 
             "user_scores": cache.get("user_scores").get(self.request.user.id), 
        })
```

backend/polls/views.py
```diff
@@ -201,25 +182,21 @@ def statistics(self, *args, **kwargs):
             time_to_next_poll = (tomorrow - datetime.now()).seconds
 
             # Setting up cache to avoid recomputing the dicts
-            cache.set("participations", participations_by_user, time_to_next_poll)
-            cache.set("victories_ranking", victories_by_user, time_to_next_poll)
-            cache.set("defeats_ranking", defeats_by_user, time_to_next_poll)
+            cache.set("poll_leaderboard", poll_leaderboard, time_to_next_poll)
 
         sorted_users_participations = sorted(
```
abocquet, line 189: Could we avoid to  re-sort it every time?

backend/polls/views.py
```diff
@@ -26,6 +26,35 @@
 )
 
 
+def generate_stats():
+    # Computing the rankings...
+    all_users = User.objects.all()
+    all_polls = Poll.objects.all()
```
abocquet, line 32:
```suggestion
    all_polls = Poll.objects.filter(has_been_published=True).all()
```

backend/polls/views.py
```diff
@@ -26,6 +26,35 @@
...
+    poll_leaderboard = {}
+
+    for user in all_users:
+        poll_leaderboard[user.id] = [0, 0, 0]
+
+    for poll in all_polls:
+        if not poll.has_been_published:
+            continue
```
abocquet, line 40:
```suggestion
```

backend/polls/views.py
```diff
@@ -201,25 +182,21 @@ def statistics(self, *args, **kwargs):
             time_to_next_poll = (tomorrow - datetime.now()).seconds
```
abocquet, line 184: Instead of adding 1 day (24 hours), it should select the next day at 00:00. Currently I'm not sure the cache is renewed often enough

abocquet (changes requested):
Way better than before, only a few more changes to do

backend/polls/views.py
victor-amblard: I think that's already what I'm doing (see previous lines). Tomorrow is day + 1 at 12am.

backend/polls/views.py
victor-amblard: I think it's not a big deal because there are at most ~5k users so it should take around 1ms to sort the list.
What would you suggest?

backend/polls/views.py
abocquet: Indeed, my bad

backend/polls/views.py
abocquet: Ok let's leave it that way

abocquet (approved):

PR #316, opened 2021-09-05
Bug fixes + NEW Calendar widget

delmaass:
## Bug fixes :

- The Event filter returned an error when the queryset was empty, because of the order_by method on the annotated queryset. I just added a condition on the annotated queryset : if it is empty, it returns it without trying to order. 
anonymines_no_events

- The locale time was not fixed to "fr" in the DayTimePickerInputField
- The Birthday widget was using the method "getDay" instead of "getDate"
- The Birthday widget was not getting the correct month (January = 0 so we have to get the month minus 1)

## NEW : Calendar widget

It displays the events to come during the next two weeks on the homepage.
You can find a visual attached to this post. It is pretty the same as the Birthday widget.

calendar_widget

home_with_calendar

anonymines_events

I have also written some tests on backside to make sure the events displayed are the correct one.


frontend/src/components/widgets/CalendarWidget.tsx
```diff
@@ -0,0 +1,93 @@
...
+                                .format("dddd DD MMMM")} ${endsAt.getHours()}:${endsAt.getMinutes()}`}
+                          </span>
+                        </div>
+                      </div>
+                  )}) }
+                </Row>
+              </div>
+            );
+          })
+      ) : "Aucun évènement à venir dans les 2 prochaines semaines :(" }
```
abocquet, line 89: Same

frontend/src/components/widgets/CalendarWidget.tsx
```diff
@@ -0,0 +1,93 @@
...
+export const CalendarWidget = ({ ...props }) => {
+  const { data, error, status } = useBetterQuery<any>(
+    ["subscriptions.calendar.get"],
+    api.subscriptions.calendar,
+    { refetchOnWindowFocus: false }
+  );
+
+  return (
+    <Widget
+      name={"Evènements à venir"}
```
abocquet, line 24: I guess you meant "Anniversaires"

abocquet (changes requested):
LGTM, 2 fix before we can approve:
- cf comments
- I see you committed new files but they are never used in fixtures. Please either remove them, or add them in fixtures so people see them show up when they reset the DB

delmaass:
Thanks for the review!
If I well understood, I had to delete the extra media files. Also, I did the necesary about the workflows for their approval.

abocquet (approved):

PR #324, opened 2022-04-01
Alexiane/associations

alexianelaude:
- fix a few bugs in association edition
- Hide associations from first years
-

### DIFF
--- a/backend/associations/views/association.py
+++ b/backend/associations/views/association.py
@@ -135,15 +135,11 @@ class AssociationViewSet(viewsets.ModelViewSet):
             return AssociationSerializer
 
     def list(self, request, *args, **kwargs):
-        print("listing assoc")
         queryset = self.filter_queryset(self.get_queryset())
         user = User.objects.get(pk=request.user.id)
 
-        queryset = [
-            a
-            for a in queryset
-            if not (a.is_hidden and user.is_in_first_year)
-        ]
+        if user.is_in_first_year:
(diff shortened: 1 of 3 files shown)


backend/associations/views/association.py
```diff
@@ -133,6 +134,25 @@ def get_serializer_class(self):
         else:
             return AssociationSerializer
 
+    def list(self, request, *args, **kwargs):
+        print("listing assoc")
```
abocquet, line 138: you may want to remove this line

backend/associations/views/association.py
```diff
@@ -133,6 +134,25 @@ def get_serializer_class(self):
...
+        queryset = self.filter_queryset(self.get_queryset())
+        user = User.objects.get(pk=request.user.id)
+
+        queryset = [
+            a
+            for a in queryset
+            if not (a.is_hidden and user.is_in_first_year)
+        ]
```
abocquet, line 146:
```
if user.is_in_first_year:
    queryset = queryset.filter(is_hidden=False)
```
would leave the filtering on the SQL side and be more efficient (especially for the pagination afterwards)

I'm not sure on the syntax here but you can check https://docs.djangoproject.com/en/4.0/ref/models/querysets/#filter

abocquet (changes requested):
It's awesome to see people pick up the project!
It looks good to me, there is a small change to do but it should be quick

abocquet (approved):

PR #325, opened 2022-04-01
Alexiane genealogy

alexianelaude:
- Find closest parent and display relation tree (pretty ugly for now...)

Screenshot 2022-04-01 at 4 59 16 PM

TODO: make this display prettier? (with an actual tree) + should we include the roommate relationship?

### DIFF
--- a/backend/authentication/views/genealogy.py
+++ b/backend/authentication/views/genealogy.py
@@ -5,21 +5,14 @@ from authentication.models import User
 
 class GenealogyView(views.APIView):
     @staticmethod
-    def chemin_to_html(chemin):
+    def chemin_links(chemin):
         """
-            Effectue un rendu en code HTML d'une liste d'élèves en précisant les
-            relations entre deux élèves successifs.
+            For a given chemin of n students, returns list of size n-1 of links between each students
         """
-        if chemin:
-            chemin_string = '<a href = "' + chemin[0].get_absolute_url() + '">' + chemin[0].first_name + ' ' + chemin[
-                0].last_name + '</a>'
-            for i in range(len(chemin) - 1):
-                chemin_string = chemin_string + ', ' + chemin[i].relation_with(chemin[i + 1]) + ' de ' + '<a href = "' + \
-                                chemin[i + 1].get_absolute_url() + '">' + chemin[i + 1].first_name + ' ' + chemin[
-                                    i + 1].last_name + '</a>'
-        else:
-            chemin_string = "Aucun chemin existant"
-        return chemin_string
+        chemin_links = []
+        for i in range(len(chemin) - 1):
+            chemin_links.append(chemin[i].relation_with(chemin[i + 1]) + ' de ')
+        return chemin_links
 
     def get(self, request):
         eleves = User.objects.all()
@@ -33,5 +26,8 @@ class GenealogyView(views.APIView):
         start = User.objects.get(id=request.data['start_username']['value'])
         end = User.objects.get(id=request.data['end_username']['value'])
         result = User.BFS(start, end)
-        result_string: str = self.chemin_to_html(result)
-        return Response({'result_string': result_string})
+        if result is None:
+            return Response({'linkFound': False, 'result': [], 'resultLinks': []})
+        result_links = self.chemin_links(result)
+        result = [user.id for user in result]
+        return Response({'linkFound': True, 'result': result, 'resultLinks': result_links})
--- a/frontend/src/components/user/trombi/Genealogy.tsx
+++ b/frontend/src/components/user/trombi/Genealogy.tsx
@@ -1,4 +1,4 @@
-import React, {useContext} from "react";
+import React, {useContext, useState} from "react";
 import Col from "react-bootstrap/Col";
 import Row from "react-bootstrap/Row";
 import Container from "react-bootstrap/Container";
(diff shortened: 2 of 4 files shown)


backend/authentication/models/user.py
```diff
@@ -192,3 +194,55 @@ def get_associations_with_permission(self, permission: str) -> set:
...
+            node = path[-1]
+            if node == end:
+                return path
+            if not node.id in visited:
+                visited.append(node.id)
+                for adjacent in node.genealogy_relations():
+                    new_path = list(path)
+                    new_path.append(adjacent)
+                    queue.append(new_path)
+        return None
```
abocquet, line 248: This method should be a separate method in `genealogy.py` as it is not intrinsically linked to the model

backend/authentication/models/user.py
```diff
@@ -192,3 +194,55 @@ def get_associations_with_permission(self, permission: str) -> set:
...
+
+    def get_absolute_url(self):
+        return '/profils/'+self.id
+
+    ## For students' genealogy ##
+
+    def genealogy_relations(self):
+        """All neighbors of a student in the Minor's graph"""
+        successors = []
+        if self.roommate.all:
```
abocquet, line 206: I believe `all` is a function here, so if that is the case, I'm not sure what we are testing here

backend/authentication/models/user.py
```diff
@@ -119,6 +119,8 @@ class User(AbstractBaseUser, PermissionsMixin):
     minesparent = models.ManyToManyField(
         "self", related_name="fillots", symmetrical=False, default=None
     )
+    cousinast = models.ManyToManyField("self", symmetrical=True, default=None)
```
abocquet, line 122:
```suggestion
    ast_cousin = models.ManyToManyField("self", symmetrical=True, default=None)
```

frontend/src/components/user/profile/ProfileRelated.tsx
```diff
@@ -17,10 +17,15 @@ const mapToUserAvatar = (userList) =>
...
 
+const hasMinesParent = (profile: Profile) => (profile.minesparent && profile.minesparent.length > 0);
+const hasFillot = (profile: Profile) => (profile.fillots && profile.fillots.length > 0);
+const hasCousinAst = (profile: Profile) => (profile.cousinast && profile.cousinast.length > 0);
+
 export const ProfileRelated = ({ profile }: { profile: Profile }) =>
-  profile.roommate.length > 0 ||
-  profile.minesparent.length > 0 ||
-  profile.fillots.length > 0 ? (
+    profile.roommate.length > 0 ||
```
abocquet, line 25: Why don't we have a `hasRoomate` function?

frontend/src/components/user/profile/ProfileRelated.tsx
```diff
@@ -17,10 +17,15 @@ const mapToUserAvatar = (userList) =>
     />
   ));
 
+const hasMinesParent = (profile: Profile) => (profile.minesparent && profile.minesparent.length > 0);
+const hasFillot = (profile: Profile) => (profile.fillots && profile.fillots.length > 0);
+const hasCousinAst = (profile: Profile) => (profile.cousinast && profile.cousinast.length > 0);
```
abocquet, line 22: This makes the code easier the read 👍

backend/authentication/views/genealogy.py
```diff
@@ -0,0 +1,40 @@
...
+        for i in range(len(chemin) - 1):
+            chemin_links.append(chemin[i].relation_with(chemin[i + 1]) + ' de ')
+        return chemin_links
+
+    def get(self, request):
+        eleves = User.objects.all()
+        result = []
+        recherche = False
+        result_string = ""
+        return Response({'eleves': eleves, 'result': result, 'result_string': result_string, 'recherche': recherche})
```
abocquet, line 23: we probably want to avoid returning the list of all students at once, especially if there is search in the request

backend/authentication/views/genealogy.py
```diff
@@ -0,0 +1,40 @@
...
+
+    def post(self, request):
```
abocquet, line 25: If it gives an information and does not modify anything, if should be a `GET`method. In `api/users.ts`, having `getGenealogy` calling `api.post` is a bit weird. Instead you can simply declare a new path with a direct method, and you'll spare the dedicated `APIView`

backend/authentication/models/user.py
```diff
@@ -192,3 +194,55 @@ def get_associations_with_permission(self, permission: str) -> set:
         from associations.models.role import Role
 
         return set(Role.objects.filter(query).values_list("association__pk", flat=True))
+
+    def get_absolute_url(self):
+        return '/profils/'+self.id
```
abocquet, line 199: I don't see this function called anywhere

abocquet (changes requested):
Thank you for this addition, it will be useful!

This PR is a bit long, generally we try to keep them under 150 lines (sometimes more if there is UI). To have more precise feedback, could you split your PR please? Suggestion:
1/ introducing the AST in the model + basic view (profile, ...)
2/ declare the `/genealogy` views in the backend with the more advanced functionnalities
3/ add the frontend
I'll give you feedback ASAP each time so you can move forward.

We try to keep the whole repo in english and avoid "frenglish", as it leads to confusion. In this regard, `ast_cousin` would be fit than 'cousinast'. For "minesparent" it's a pun with the "godparent" word.

In order to make the pipelines pass, you should reformat your code with "black" (https://github.com/psf/black), you can either reformat the whole codebase with it at once or reformat files as you save them (there is a PyCharm extension). This is important so there never are changes linked to formatting in PR reviews, which is very noisy for nothing.

You did a great job, the few comments should be quick to fix 👍

backend/authentication/models/user.py
```diff
@@ -192,3 +194,55 @@ def get_associations_with_permission(self, permission: str) -> set:
...
+        if self in student.fillots.all():
+            return "fillot.e"
+        if self in student.cousinast.all():
+            return "cousin.e"
+        return None
+
+    @staticmethod
+    def BFS(start, end):
+        """Breadth-First-Search algorithm, to find the shortest path between two students"""
+        visited = []
```
abocquet, line 234: using a `set` will provide `O(log n)` complexity vs `O(n)` with a plain array in the `in` in line 242

backend/authentication/models/user.py
```diff
@@ -192,3 +194,55 @@ def get_associations_with_permission(self, permission: str) -> set:
...
+        queue = []
+        queue.append([start])
```
abocquet, line 236:
```suggestion
        queue = [[start]]
```

PR #326, opened 2022-04-02
Game features: list and retrieve games

delmaass:
This is the first PR of the creation of the "game framework" which have the ambition to allow anyone to create and add games easily to the "portail des élèves".

This framework should:
- allow anyone to create and add games simply by adding the game to the list and by creating the appropriate container and server for the game (e.g. for multiplayer games) ;
- have a global score system : a game server only have to send a request to the API with the value of the score and the user ;
- display a leaderboard according to the users' scores.

Here is some screenshots of the final result with an example of a functional game (Pictionary, a multiplayer game) : 
Screenshot 2022-04-02 at 16-35-43 React App
Screenshot 2022-04-02 at 16-36-48 React App

This PR only affects the backend with the creation of a new app `games` and the addition of the "list and retrieve games" feature.

Note : you can see a lot of line diff, actually I only edit the `backend/games/` folder (small modifications outside this folder - e.g. on `settings.py`, no modifications made on the frontend)

Many thanks for the review :)


delmaass:
The PR #328 solves the fail of Django CI. You can run `python manage.py test games` locally to ensure that the tests are functional.

abocquet:
The PR seems OK but it modifies excessively too many lines in the code that are not related to the PR (as you said yourself). I don't know if your code editor chose to delete so many files and change file formatting, but this behavior should be disabled.

In this state we cannot merge this PR, as it would probably cause some mayhem in the future 
- merge conflicts because of bad formatting
- files not present when they should
- someone else with another formatter that changes the files back to their original state. This would create PRs of thousands of lines for nothing each time

Also, version bumps should probably have their own PR, or at least the upgraded packages should be linked to the PR
Please note that we chose to share `package-lock.json`, according nom guidelines https://docs.npmjs.com/cli/v6/configuring-npm/package-locks#using-locked-packages

PR #333, opened 2022-04-03
Games: list scores (& create if staff user or game service) + CLI

delmaass:
This PR introduces the Score system!
Each time you play a game, the associated game system should create a Score entry for you.

Backend :
- List scores and filter by game ;
- Possibility to create a score if you are a staff user or if a game system uses the auth token of a staff user.


backend/games/models/scores.py
```diff
@@ -0,0 +1,22 @@
...
+
+    user = models.ForeignKey(
+        User, verbose_name="user", on_delete=models.CASCADE, related_name="score"
+    )
+
+    score = models.IntegerField(
+        verbose_name="score", validators=[MinValueValidator(0)], default=0
+    )
+
+    when = models.DateTimeField(verbose_name="when", auto_now_add=True, blank=True)
```
abocquet, line 22: Maybe `created_on` would be easier to follow, as it's the term usually used in the codebase

backend/games/views/scores.py
```diff
@@ -0,0 +1,59 @@
...
+class ActionBasedPermission(AllowAny):
+    """
+    Grant or deny access to a view, based on a mapping in view.action_permissions
+    """
+
+    def has_permission(self, request, view):
+        for klass, actions in getattr(view, "action_permissions", {}).items():
+            if view.action in actions:
+                return klass().has_permission(request, view)
+        return False
```
abocquet, line 22: Could you explain what this does? It seems very generic so I'm surprised we have to reimplement it here

abocquet (approved):

backend/games/views/scores.py
delmaass: It allows a viewset to have a specific permission class for each allowed action. Here I needed to have the permission class IsAdminUser for the CREATE action, and the permission class IsAuthenticated for the LIST action. (I found this solution on stackoverflow, didn't find better...)

abocquet (approved):

PR #336, opened 2022-04-04
Games: Leaderboard

delmaass:
This PR introduces the Leaderboard.
It is simply the sum of the Scores' value for each user, sorted in descending order.
You can filter the leaderboard by a Game, to display it next to a Game container.

### DIFF
--- a/backend/games/tests/test_leaderboard.py
+++ b/backend/games/tests/test_leaderboard.py
@@ -1,0 +1,79 @@
+from datetime import datetime
+from games.models.scores import Score
+from backend.tests_utils import WeakAuthenticationBaseTestCase
+from django.db.models import Sum
+
+
+class LeaderboardTestCase(WeakAuthenticationBaseTestCase):
+    fixtures = ["test_authentication.yaml", "games.yaml", "test_scores.yaml"]
+
+    EXPECTED_FIELDS = {"user", "total_score"}
+
+    def endpoint_list(self):
+        return "/games/scores/leaderboard/"
+
+    def list(self):
+        return self.get(self.endpoint_list())
+
+    def endpoint_filter_game(self, game):
+        return f"/games/scores/leaderboard/?game={game}"
+
+    def filter_game(self, game):
+        return self.get(self.endpoint_filter_game(game))
+
+    def validate_leaderboard_score(self, prev_score, users, score, game=None):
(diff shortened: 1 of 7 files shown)


backend/games/tests/test_leaderboard.py
```diff
@@ -0,0 +1,79 @@
...
+    ########
+    # LIST #
+    ########
+
+    def test_if_not_logged_in_then_cannot_list(self):
+        res = self.list()
+        self.assertStatusCode(res, 401)
+
+    def if_logged_in_then_can_list(self):
+        print(self.login("17simple"))
```
abocquet, line 53: This might not be supposed to be here

backend/games/tests/test_leaderboard.py
```diff
@@ -0,0 +1,79 @@
...
+
+class LeaderboardTestCase(WeakAuthenticationBaseTestCase):
+    fixtures = ["test_authentication.yaml", "games.yaml", "test_scores.yaml"]
+
+    EXPECTED_FIELDS = {"user", "total_score"}
+
+    def endpoint_list(self):
+        return "/games/scores/leaderboard/"
+
+    def list(self):
```
abocquet, line 15: `list` is a very generic term, it would be easier to read with something like `query_list_endpoint`

backend/games/tests/test_leaderboard.py
```diff
@@ -0,0 +1,79 @@
+from datetime import datetime
+from games.models.scores import Score
+from backend.tests_utils import WeakAuthenticationBaseTestCase
+from django.db.models import Sum
+
+
+class LeaderboardTestCase(WeakAuthenticationBaseTestCase):
```
abocquet, line 7: Should we add a test that tries to submit a negative score?

abocquet (approved):
Some nits in the tests, otherwise LGTM

backend/games/tests/test_leaderboard.py
delmaass: Are you talking about the print function ?

backend/games/tests/test_leaderboard.py
delmaass: U are right, it should be tested 👍 I can add this test in `test_scores.py`, in another PR.

PR #355, opened 2022-04-08
Fix marketplace and library update/creation when updating association

alexianelaude:
- On updating an association, the current checkboxes to enable/disable marketplace and library are not working.
- If the marketplace/library does not yet exists, it is created
- Otherwise, it is updated

### DIFF
--- a/frontend/src/components/associations/settings/AssociationOptInSettings.tsx
+++ b/frontend/src/components/associations/settings/AssociationOptInSettings.tsx
@@ -20,7 +20,7 @@ export const AssociationOptInSettings = ({ association }) => (
           <EnableModuleForm
             id={association.id}
             label={"Magasin"}
-            apiMethod={api.marketplace.update}
+            apiMethod={api.marketplace.createOrUpdate}
             initialValue={association.enabledModules.includes("marketplace")}
           />
         </Col>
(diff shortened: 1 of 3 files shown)


frontend/src/components/associations/settings/AssociationOptInSettings.tsx
```diff
@@ -69,15 +69,14 @@ const EnableModuleForm = ({
...
-        save({ id: id, enabled: !checked });
-      }}
-    >
+    <label>
       <input
         type="checkbox"
         className="custom-switch-input"
         checked={checked}
+        onChange={() => {
+          save({ id: id, enabled: !checked, association: id, products: [] });
```
abocquet, line 78: This will reset the list of product each time right? This may be dangerous in case of misclick

abocquet (approved):

frontend/src/components/associations/settings/AssociationOptInSettings.tsx
alexianelaude: The products field will only be used if the marketplace doesn't exist yet, in which case it will be created with an empty products list. In case of an update, the 'products' value will not be used as the update method from the MarketplaceViewSet only changes the 'enabled' boolean.

abocquet: So if we always initialize the product list to an empty list, why do we specify it at all?

